PR #42, opened 2022-03-18
Move interrogate docstring coverage settings to pyproject.toml

karlhigley:
The docstring coverage pre-commit hooks has been blocking some commits that touch tests, since the tests weren't properly excluded. Putting the configuration in `pyproject.toml` (instead of in the command) gives a clean way to fix that.

### DIFF
--- a/.github/workflows/cpu-ci.yml
+++ b/.github/workflows/cpu-ci.yml
@@ -41,7 +41,7 @@ jobs:
         isort -c .
     - name: Lint with interrogate
       run: |
-        interrogate -viImMpPsSCn -f 35.0 merlin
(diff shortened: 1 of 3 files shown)


.github/workflows/cpu-ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         isort -c .
     - name: Lint with interrogate
       run: |
-        interrogate -viImMpPsSCn -f 35.0 merlin
+        interrogate --config=pyproject.toml
```
benfred, line 44: nice!

benfred (approved):

PR #49, opened 2022-03-25
chore: linting

mikemckiernan:
Check me...I modified setup.cfg because I was otherwise
receiving the foll error:

```text
merlin/systems/triton/model_config_pb2.py:24: SpecificH ==> specify, specific
```

CI will prolly reveal it soon, but I'll try to fess up before it show it.  I had to use the foll to commit:

```shell
SKIP=interrogate git commit
```

### DIFF
--- a/merlin/systems/dag/ensemble.py
+++ b/merlin/systems/dag/ensemble.py
@@ -21,9 +21,9 @@ from merlin.dag import postorder_iter_nodes
 os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
 
 from google.protobuf import text_format  # noqa
+from merlin.dag import Graph  # noqa
 
 import merlin.systems.triton.model_config_pb2 as model_config  # noqa
-from merlin.dag import Graph  # noqa
 from merlin.systems.triton.export import _convert_dtype  # noqa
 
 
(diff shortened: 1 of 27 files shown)


merlin/systems/dag/ensemble.py
```diff
@@ -21,9 +21,9 @@
 os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
 
 from google.protobuf import text_format  # noqa
+from merlin.dag import Graph  # noqa
```
benfred, line 24: I think we should be adding merlin as a known first party package for isort (so it shouldn't be grouped with thirdparty things like google.protobuf).

something like https://github.com/NVIDIA-Merlin/models/pull/216

benfred: (involves adding this line to the isort config https://github.com/NVIDIA-Merlin/models/blob/303a91eebb4859c7d4c5065f345b33e45962c69e/pyproject.toml#L21  )

karlhigley: Added

benfred (approved):

benfred:
This change looks good to me - but give me a couple minutes before merging, since I'm curious why the github actions CI didn't kick off here

benfred:
Looks like because of 
```actions/download-artifact@v2, fnkr/github-action-ghr@v1.3, actions/setup-python@v2, actions/checkout@v2, and actions/upload-artifact@v2 are not allowed to be used in NVIDIA-Merlin/systems. Actions in this workflow must be: within a repository owned by NVIDIA-Merlin.
```
- I'm taking a look at fixing

PR #48, opened 2022-03-24
Merlin Systems example notebook

jperez999:
This PR introduces a basic example notebook for running a quick workflow >> model ensemble. It is not complete it still requires changes to the tritonserver request launch section but placeholders have been added for the code necessary. Put this up now to get comments on writing style and content.

PR #50, opened 2022-03-26
Add a README

karlhigley:
Depends on example code from https://github.com/NVIDIA-Merlin/systems/pull/48 — either we'll copy example code from that notebook, or we'll link to it, or both


karlhigley:
I'd kind of like to show how to build a simple four-stage recommender in the README. I wonder if we might be able to use the POC test as a source for example code @jperez999 ?

PR #61, opened 2022-03-30
add torch to cpu ci

jperez999:
adds torch package to cpu ci runs for better coverage of unit tests.


karlhigley:
Curious: what additional test coverage does this unlock? (Wondering since I didn't think we had Torch support in this library yet.)

jperez999:
https://github.com/NVIDIA-Merlin/systems/blob/ccd0e15941959f46b7152a7dbda07f7b90ce713a/tests/unit/systems/test_export.py#L31
https://github.com/NVIDIA-Merlin/systems/blob/1b5fea25d8738ea37fca3a73e06963414a2af294/tests/unit/systems/inference_utils.py#L20

karlhigley (approved):
Since we're not actually using PyTorch yet, I think we could also remove it, but this is easier for now 🤷🏻

PR #65, opened 2022-03-30
Bump docstring coverage check threshold to 70%

karlhigley:


EvenOldridge (approved):
Awesome that we can bump it by this much after the review.  :-)

PR #77, opened 2022-04-03
docs: Add nightly build and start API

mikemckiernan:
* Took a few guesses at the API.  I need
  some guidance.


mikemckiernan:
I'll be the bot for now.  PTAL.

<https://nvidia-merlin.github.io/systems/review/pr-77/>

docs/source/api.rst
```diff
@@ -0,0 +1,60 @@
+*****************
+API Documentation
+*****************
+
+.. currentmodule:: merlin.systems
+
+
+Workflow Graph Constructors
```
karlhigley, line 8: `Workflow` here should be `Ensemble`—which are very similar, so you have the right general idea, but this aligns the terminology with Triton.

docs/source/api.rst
```diff
@@ -0,0 +1,60 @@
...
+---------------------------
+
+.. currentmodule:: merlin.systems.dag
+
+.. autosummary::
+   :toctree: generated
+
+   Ensemble
+   Node
```
karlhigley, line 17: We don't expect that users will need to interact with `Node` directly, so not sure if it should be included here or not. No harm, I suppose.

docs/source/api.rst
```diff
@@ -0,0 +1,60 @@
...
+   tensorflow.PredictTensorflow
+   softmax_sampling.SoftmaxSampling
+   session_filter.FilterCandidates
+   unroll_features.UnrollFeatures
+
+.. faiss.QueryFaiss
+.. feast.QueryFeast
+
+
+Workflow Functions
```
karlhigley, line 38: From here down, I don't think we expect users to interact with any of these directly. (We should probably make them private.)

karlhigley (approved):

PR #84, opened 2022-04-07
Get feast repo path from feature store

jperez999:
This PR removes an unnecessary param from one of the creation mechanism for the feast operator. The path is not necessary if you have a feature store already created, because that object has a reference to the path.


karlhigley (approved):

jperez999:
This requires updates in the POC example, should we coordinate before updating? @rnyak @karlhigley

benfred (approved):

PR #91, opened 2022-04-27
Update README.rst

bschifferer:
I think the README should point to the examples of Merlin Systems.


mikemckiernan (commented):
The update is accurate, but we already merged a PR that A) uses the correct path and B) is in README.md rather than README.rst.

PR #109, opened 2022-05-24
add unit tests script

jperez999:
adds test script to keep systems in line with other merlin repos. This now does quite a bit more and requires the we first merge this pr in the nvtabular repo: https://github.com/NVIDIA-Merlin/NVTabular/pull/1563. With that change then we can safely import all the messages and enums from a file that interacts with the google protobuf internal database that loads files when they are compiled via protoc.


karlhigley:
rerun tests

jperez999:
rerun tests

karlhigley:
It looks like this hit issues and I think I saw a different fix for this go by in NVT. Do we still need this PR? (Fine if we do, just not sure.)

jperez999:
rerun tests

jperez999:
> It looks like this hit issues and I think I saw a different fix for this go by in NVT. Do we still need this PR? (Fine if we do, just not sure.)

yes we need it, but it has been revamped. we have decoupled the need of model config from nvtabular and solely rely on tritonclient package version.

karlhigley (approved):

PR #110, opened 2022-05-26
Add an Operator for Forest Inference Libary (FIL) Triton backend.

oliverholworthy:
Add an Operator for Forest Inference Libary (FIL) Triton backend.

- [x] figure out output signature automatically (e.g. classification, + with proba)
- [x] allow params to be passed in and take precedence over defaults

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@
 import os
 import sys
 
-from setuptools import find_namespace_packages, find_packages, setup
+from setuptools import find_namespace_packages, setup
 
 try:
     import versioneer
@@ -40,7 +40,7 @@ install_reqs = parse_requirements("./requirements.txt")
 setup(
     name="merlin-systems",
     version=versioneer.get_version(),
-    packages=find_packages() + find_namespace_packages(),
+    packages=find_namespace_packages(include=["merlin*"]),
     url="https://github.com/NVIDIA-Merlin/systems",
     author="NVIDIA Corporation",
     license="Apache 2.0",


merlin/systems/dag/ops/fil.py
```diff
@@ -0,0 +1,91 @@
+import pathlib
+
+import numpy as np
+from google.protobuf import text_format  # noqa
+
+import merlin.systems.triton.model_config_pb2 as model_config  # noqa
```
karlhigley, line 6: In parallel with your PR, we just switched over to importing this from the tritonclient in #109 to avoid some issues with building/registering classes from Protobuf twice:
```suggestion
import tritonclient.grpc.model_config_pb2 as model_config # noqa
```

merlin/systems/dag/ops/fil.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, model):
+        """Instantiate a FILPredict inference operator.
+
+        Parameters
+        ----------
+        model : xgboost.Booster
+            An XGBoost model.
+        """
+        self.model = model
+        self.input_schema = Schema([ColumnSchema("input__0", dtype=np.int32)])
```
karlhigley, line 27: Since we have methods to compute the schemas, might make sense to return these schemas from those methods (even though they're functionally constants in this case.)

merlin/systems/dag/ops/fil.py
```diff
@@ -0,0 +1,91 @@
...
+        self.output_schema = Schema([ColumnSchema("output__0", dtype=np.int32)])
```
karlhigley, line 28: Wondering if this should be a float type like `np.float32` but not completely sure what FIL expects.

tests/unit/systems/test_fil.py
```diff
@@ -0,0 +1,102 @@
...
+import merlin.systems.dag.ops.fil as fil_op
+from merlin.dag import ColumnSelector, Graph
+from merlin.schema import Schema
+
+# this needs to be before any modules that import protobuf
+os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
+
+from google.protobuf import text_format  # noqa
+
+model_config = pytest.importorskip("nvtabular.inference.triton.model_config_pb2")
```
karlhigley, line 32:
```suggestion
model_config = pytest.importorskip("tritonclient.grpc.model_config_pb2")
```

.github/workflows/cpu-ci.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         python -m pip install --upgrade pip setuptools==59.4.0 wheel 
         python -m pip install -r requirements-dev.txt
         python -m pip install tensorflow
+        python -m pip install xgboost
```
karlhigley, line 36: Now that we're supporting more than one modeling library, we should make `tensorflow` and `xgboost` optional dependencies. For this PR, probably good enough to make `xgboost` optional and we do the same for TF in a separate PR.

karlhigley (commented):
Looks good! Left a few minor comments

.github/workflows/cpu-ci.yml
oliverholworthy: can you clarify what you mean by optional in the context of this line. The systems package currently only has two dependencies, merlin-core and nvtabular. Do you mean adding something like we have with models where we can pip install merlin-models[xgboost] ? 

the FIL op as currently written doesn't strictly depend on xgboost (import-wise), only in the sense that some calling it will need to load a model - maybe that could be left up to them to install?

merlin/systems/dag/ops/fil.py
oliverholworthy: yes, FIL does expect float32

merlin/systems/dag/ops/fil.py
oliverholworthy: that is a bit cleaner, updated

.github/workflows/cpu-ci.yml
karlhigley: Yeah, I think we might want to set up `pip install merlin-systems[xgboost]`, but it's out of scope for this PR. This line just reminded me that it's a thing to do.

PR #112, opened 2022-06-02
Fix broken PoC notebook due mismatching output names between config file and saved ranking model

rnyak:
Currently,  [02-Deploying-multi-stage-RecSys-with-Merlin-Systems.ipynb](https://github.com/NVIDIA-Merlin/Merlin/blob/main/examples/Building-and-deploying-multi-stage-RecSys/02-Deploying-multi-stage-RecSys-with-Merlin-Systems.ipynb) nb is broken, due to ranking model output name mismatch between the config files, and saved model.

we get the following errors:

`ValueError: Missing columns ['output_1'] found in operatorSubsetColumns during compute_input_schema.`

This can be fixed by setting the proper output name `click/click/binary_classification_task` in  the following lines:

```
ordering = combined_features["item_id"] >> SoftmaxSampling(
    relevance_col=ranking["click"], topk=top_k, temperature=20.0
)
```

However, this is not enough. Triton also complains about output names, therefoer cannot load `5_predicttensorflow` model.

This PR proposes a solution to that issue.

### DIFF
--- a/merlin/systems/dag/ops/tensorflow.py
+++ b/merlin/systems/dag/ops/tensorflow.py
@@ -79,7 +79,7 @@ class PredictTensorflow(InferenceOperator):
         outputs = list(default_signature.structured_outputs.values())
 
         input_col_names = [col.name.split("/")[0] for col in inputs]
-        output_col_names = [col.name.split("/")[0] for col in outputs]
+        output_col_names = [col.name for col in outputs]
 
         self.input_schema = Schema()
         for col, input_col in zip(input_col_names, inputs):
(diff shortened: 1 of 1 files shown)


benfred (commented):
Do you have an example of the output column names that will trigger this? Also Can we add a unittest that will test this out?

rnyak:
> Do you have an example of the output column names that will trigger this? Also Can we add a unittest that will test this out?

yes the unit test is already available, it was just not in the correct folder. This PR https://github.com/NVIDIA-Merlin/Merlin/pull/364 pushes it to the `unit` folder.

The output column name that will trigger this is `click/binary_classification_task` and it was `output_1` before.

PR #125, opened 2022-06-21
Add hugectr op

jperez999:
This PR addresses concerns around hugectr support. This represents the foundations for supporting a hugectr model in merlin systems.

PR #129, opened 2022-07-06
Hugectr base

jperez999:
This PR will introduce the initial hugectr Operator. This operator works along and will need a wrapper operator to handle inputs coming from a dataframe. The PR lays the foundation for using Hugectr in systems. Allows you to pass a model or path to a model and it is loaded, relevant information extracted and the necessary artifacts are created (ps.json, model files, model.json, config.pbtxt for triton).


jperez999:
rerun tests

merlin/systems/dag/ops/hugectr.py
```diff
@@ -0,0 +1,300 @@
...
+            if "opt" not in path.name
+        ][0]
+        sparse_pattern = "*_sparse_*.model"
+        sparse_paths = [
+            os.path.join(hugectr_model_path, path.name)
+            for path in hugectr_model_path.glob(sparse_pattern)
+            if "opt" not in path.name
+        ]
+
+        config_dict = dict()
```
karlhigley, line 168: I think the linter suggestion for this is to use `{}` instead of `dict()`

merlin/systems/dag/ops/hugectr.py
```diff
@@ -0,0 +1,300 @@
...
+        model["default_value_for_each_table"] = [0.0] * len(sparse_layers)
+        model["hit_rate_threshold"] = 0.9
+        model["gpucacheper"] = self.hugectr_params["gpucacheper"]
+        model["gpucache"] = True
+        model["cache_refresh_percentage_per_iteration"] = 0.2
+        model["maxnum_catfeature_query_per_table_per_sample"] = [
+            len(x["sparse_embedding_hparam"]["slot_size_array"]) for x in sparse_layers
+        ]
+        model["embedding_vecsize_per_table"] = vec_size
+        model["embedding_table_names"] = [x["top"] for x in sparse_layers]
```
karlhigley, line 200: Wonder if might be worthwhile to extract a helper function to construct this dictionary

merlin/systems/dag/ops/hugectr.py
```diff
@@ -0,0 +1,300 @@
...
+
+        config = _hugectr_config(node_name, self.hugectr_params, max_batch_size=self.max_batch_size)
+
+        with open(os.path.join(node_export_path, "config.pbtxt"), "w") as o:
+            text_format.PrintMessage(config, o)
+
+        return config
+
+
+def _hugectr_config(name, hugectr_params, max_batch_size=None):
```
karlhigley, line 224: It seems like there's a fair amount of repetition in this method. Maybe some of this can be done with a for loop?

karlhigley (approved):
Left some style suggestions, but nothing that would block this PR once the tests pass

merlin/systems/dag/ops/hugectr.py
oliverholworthy: I'm pretty sure we implemented this a for loop @jperez999  must have been during the splitting of commits from https://github.com/NVIDIA-Merlin/systems/pull/125

EvenOldridge:
@minseokl @zehuanw We're very close to HugeCTR support in Systems but we're running into this segfault.  Can you please review with the team.

PR #134, opened 2022-07-11
Add Operator for Implicit Models

oliverholworthy:
Adds an Operator for Implicit Models `PredictImplicit`.

Self-contained operator for serving implicit models.


merlin/systems/dag/op_runner.py
```diff
@@ -30,14 +33,21 @@ def __init__(self, config, repository="./", version=1, kind=""):
             op_module = importlib.import_module(module_name)
             op_class = getattr(op_module, class_name)
 
-            operator = op_class.from_config(op_config)
+            operator = op_class.from_config(
```
oliverholworthy, line 36: @karlhigley In order to load an artifact required by an operator, passing the `model_*` related params through to all ops that use the op runner here. I'm guessing this may have considered before because the existing OperatorRunner accepts a model repository and version (though currently unused). Does this seem like a reasonable thing to do?

merlin/systems/dag/ops/implicit.py
```diff
@@ -0,0 +1,119 @@
...
+        df: InferenceDataFrame
+            A pandas or cudf dataframe that this operator will work on
+
+        Returns
+        -------
+        InferenceDataFrame
+            Returns a transformed dataframe for this operator"""
+        user_id = df["user_id"][0]
+        num_to_recommend = df["n"][0]
+        user_items = None
```
oliverholworthy, line 115: could move this to the inputs if it's a common pattern for users to pass in. My understanding is that this mostly serves as a mechanism to filter out items already seen.

benfred: I think this is ok - user_items can be left out if the filter_already_liked_items is False, and recalculate_user is also False (otherwise we need to now what the items the user has liked).

merlin/systems/dag/ops/faiss.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, index_path, topk=10):
         super().__init__()
 
     @classmethod
-    def from_config(cls, config: dict) -> "QueryFaiss":
+    def from_config(cls, config: dict, **kwargs) -> "QueryFaiss":
```
benfred, line 62: Whats the benefit of adding the `kwargs` here ?

oliverholworthy: This is required with the change made in `op_runner.py`. The from_config now receives `model_repository`, `model_name`, and `model_version` keyword arguments which enables us to read files stored alongside the model config.

This QueryFaiss operator in this case relies on the config containing the full path to the index file currently. However with this extra information could be updated to read from a location relative to the model repository and model config.

It could be clearer to update this **kwargs to specify the named arguments explicitly here (and in the other from_config methods.)

merlin/systems/dag/ops/implicit.py
```diff
@@ -0,0 +1,119 @@
...
+
+from merlin.dag import ColumnSelector  # noqa
+from merlin.schema import ColumnSchema, Schema
+from merlin.systems.dag.ops.operator import InferenceDataFrame, PipelineableInferenceOperator
+
+
+class PredictImplicit(PipelineableInferenceOperator):
+    """Operator for running inference on Implicit models.."""
+
+    def __init__(self, model, **kwargs):
```
nv-alaiacano, line 28: The `PredictTensorflow` op has `model_or_path` as its param and will load the model from a path if that's what is provided. I think we should try to be consistent with the constructor functionality, and `PredictTensorflow` is probably the most mature/popular, so matching that seems like a good idea to me.

oliverholworthy: 👍 I agree that having consistency across these Predict ops is a good idea. 

However, the implicit serialization format doesn't appear to have a mechanism for knowing which model to load. 

e.g. if you have created a model and saved to a file, for example `model.npz`.

```python
import implicit

# random training data
n = 100  # number of both users and items
user_items = csr_matrix(np.random.choice([0, 1], size=n * n, p=[0.9, 0.1]).reshape(n, n))

model = implicit.bpr.BayesianPersonalizedRanking()
model.fit(user_items)
model.save("model.npz")  # save model to file
```

to call the `load` method we require knowledge of the kind of model to be loaded. There's not an `implicit.load` function that I can see (correct me if there is something I've missed @benfred )

```python
implicit.gpu.bpr.BayesianPersonalizedRanking.load("model.npz")
```

We could change the signature to `model_or_path` at a later stage if the functionality becomes available and that is our preferred option for these predict ops.

oliverholworthy:
- Re-based on top of #158

- Moved the specification of N the number of items to recommend to the Op constructor instead of as a triton input.
  - This is because as a Triton Input it didn't seem very use-friendly. The way we write the config currentlyu requires inputs to be of shape (-1, 1) which doesn't fit very well with a single integer value. Since other Ops don't seem to support this kind of runtime config at the moment, it seems reasonable to me to put this in the constructor for now. Perhaps we can consider better ways to support the kind of arguments that are not part of the input to a model for runtime optional configuration.

merlin/systems/triton/oprunner_model.py
```diff
@@ -129,14 +129,13 @@ def execute(self, requests):
...
-
     # Handle bug in Tritonserver 22.06
     # model_repository argument became path to model.py
-    if str(model_repository).endswith(".py"):
-        model_repository_path = model_repository_path.parent
-
-    return str(model_repository_path)
+    # instead of path to model directory within the model repository
+    if model_repository.endswith(".py"):
+        return str(pathlib.Path(model_repository).parent.parent.parent)
```
oliverholworthy, line 139: example:

```python
import pathlib

pathlib.Path("/tmp/my_model_repository/my_model_name/1/model.py").parent.parent.parent
# => PosixPath('/tmp/my_model_repository')
```

oliverholworthy:
rerun tests

karlhigley:
@oliverholworthy It looks like this is failing on an XGBoost test, but not I'm sure why. It can't find the ensemble, but I'm not sure if that's an issue with the test or an issue with the environment

oliverholworthy:
rerun tests

oliverholworthy:
@karlhigley triggered another run. This might be something to do with the triton isolation issue (the reason #141 can't be merged - as it is currently written)

karlhigley:
That was my guess too, probably due to the PR stampede now that CI is unblocked 👍🏻

nv-alaiacano (approved):

PR #141, opened 2022-07-13
Enable gRPC port to be specified in `run_triton_server` and check if already running

oliverholworthy:
## Goal

Improve the reliability of local dev and CI testing using triton.

## Motivation

When a  `tritonserver` process is already running on the port we would like to bind to - the tests relying on the `run_triton_server` helper fail with an unexpected error (complaining that the model is not loaded for example). 

Enabling the gRPC server to listen on a random port allows us to run multiple ensemble tests at the same time (GPU resources permitting).

## Details

This PR will:

- Add `grpc_port`, `grpc_host`, and `backend_config` parameters to `run_triton_server`
  - Default is set to `8001`. The same  as before. If 0 is provided,a random free port will be used.
- Add a check to see if a tritonserver process is already running on the desired port and raises an exception if there is one.

### DIFF
--- a/merlin/systems/triton/utils.py
+++ b/merlin/systems/triton/utils.py
@@ -14,7 +14,7 @@ TRITON_SERVER_PATH = find_executable("tritonserver")
 
 
 @contextlib.contextmanager
-def run_triton_server(modelpath):
+def run_triton_server(modelpath, grpc_host="localhost", grpc_port=8001):
     """This function starts up a Triton server instance and returns a client to it.
 
     Parameters
@@ -28,8 +28,6 @@ def run_triton_server(modelpath):
         The client connected to the Triton server.
 
     """
-    grpc_host = "localhost"
-    grpc_port = 8001
     grpc_url = f"{grpc_host}:{grpc_port}"
 
     with grpcclient.InferenceServerClient(grpc_url) as client:
--- a/merlin/systems/triton/utils.py
+++ b/merlin/systems/triton/utils.py
@@ -28,6 +28,14 @@ def run_triton_server(modelpath):
         The client connected to the Triton server.
 
     """
+    grpc_host = "localhost"
+    grpc_port = 8001
+    grpc_url = f"{grpc_host}:{grpc_port}"
+
+    with grpcclient.InferenceServerClient(grpc_url) as client:
+        if client.is_server_ready():
+            raise RuntimeError(f"Another tritonserver is already running on {grpc_url}")
+
     cmdline = [
         TRITON_SERVER_PATH,
         "--model-repository",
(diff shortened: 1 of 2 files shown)


oliverholworthy:
We seem to be getting the `Another tritonserver is already running`  error in the Jenkins tests 🤔 Is the test runner isolated from other PRs?

karlhigley:
I'm not entirely sure, but I would be surprised if it was based on some other issues I've seen crop up. @jperez999 would probably know.

oliverholworthy:
rerun tests

merlin/systems/triton/utils.py
```diff
@@ -28,6 +28,14 @@ def run_triton_server(modelpath):
         The client connected to the Triton server.
 
     """
+    grpc_host = "localhost"
+    grpc_port = 8001
```
nv-alaiacano, line 32: maybe pull these out to be parameters (with default values)

jperez999:
> We seem to be getting the `Another tritonserver is already running` error in the Jenkins tests 🤔 Is the test runner isolated from other PRs?

It is not. Therefore multiple tritonservers can be running. Also we may not want to throw an error when a tritonserver is available. Because IIRC each tritonserver instance can only run one backend  in it... i.e. I cant run a hugectr model and tf model on the same triton instance... because one of them will not have the correct backend. So I think we do want to be able to run multiple tritonserver instances... it maybe that when we fail because we see one available we may need to kick out to different set of ports. I dont know, what do you think?

oliverholworthy:
Updated based on the comments above.

- Port added as optional parameter
- Using a random free port to make collisions less likely
  - this may open up some issues around resource use if some tritonserver processes don't shutdown correctly. Something to keep an eye out for if we choose to merge this.

merlin/systems/triton/utils.py
```diff
@@ -1,20 +1,23 @@
...
 
 
 @contextlib.contextmanager
-def run_triton_server(modelpath):
+def run_triton_server(
+    modelpath, grpc_host="localhost", grpc_port=None, backend_config="tensorflow,version=2"
+):
     """This function starts up a Triton server instance and returns a client to it.
 
     Parameters
```
nv-alaiacano, line 23: Could you update the docstring to describe the new parameters and maybe a brief description of how the port selection works now?

tests/unit/examples/test_serving_an_xgboost_model_with_merlin_systems.py
```diff
@@ -34,5 +34,5 @@ def test_example_serving_xgboost(tb):
     )
     tb.execute_cell(list(range(0, 14)))
 
-    with run_triton_server("ensemble"):
+    with run_triton_server("ensemble", grpc_port=8001):
```
nv-alaiacano, line 37: Would hard-coding triton's default port number here re-introduce the issue that the PR is trying to avoid?

oliverholworthy: So the tests have now passed with port 8001 set as default which is what was running with before. I'm tempted to say let's keep the port set to 8001 and see how often we see the new exception (added in this PR) about triton already running show up. We can then change the default to 0 (for random port selection) if we see it happening too much.

nv-alaiacano (approved):

PR #146, opened 2022-07-18
Allow merlin.models.xgb.XGBoost in PredictForest

benfred:
The merlin-models XGB wrapper wasn't supported in PredictForest. This change
adds basic support for passing the MM xgb wrapper directly to PredictForest
(without requiring a dependency on MM itself).

### DIFF
--- a/merlin/systems/dag/ops/fil.py
+++ b/merlin/systems/dag/ops/fil.py
@@ -301,7 +301,7 @@ class FIL(InferenceOperator):
             **self.parameters,
         )
 
-        with open(node_export_path / "config.pbtxt", "w") as o:
+        with open(node_export_path / "config.pbtxt", "w", encoding="utf-8") as o:
             text_format.PrintMessage(config, o)
 
         return config
--- a/merlin/systems/dag/ops/fil.py
+++ b/merlin/systems/dag/ops/fil.py
@@ -384,6 +384,9 @@ def get_fil_model(model) -> FILModel:
         ),
     ):
         fil_model = SKLearnRandomForest(model)
+    elif hasattr(model, "booster"):
+        # support the merlin.models.xgb.XGBoost wrapper too
+        return get_fil_model(model.booster)
     else:
         supported_model_types = {
             "xgboost.Booster",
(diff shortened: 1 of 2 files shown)


benfred:
@radekosmulski @bschifferer  Doing something like this is what I was suggested in the docs sync earlier today

radekosmulski:
@benfred this looks great!

benfred:
CI is failing here - but will be fixed by https://github.com/NVIDIA-Merlin/systems/pull/147

merlin/systems/dag/ops/fil.py
```diff
@@ -384,6 +384,9 @@ def get_fil_model(model) -> FILModel:
         ),
     ):
         fil_model = SKLearnRandomForest(model)
+    elif hasattr(model, "booster"):
```
EvenOldridge, line 387: We should move this up with the other xgboost models in the elif.

Also not entirely clear why:

>     elif xgboost and isinstance(model, xgboost.XGBModel):
>         fil_model = XGBoost(model.get_booster())

isn't picking up this case.  Am I correct in understanding that they're both trying to get the booster from the model?  Or is this a different case.

EvenOldridge (dismissed):

merlin/systems/dag/ops/fil.py
oliverholworthy: `isinstance(model, xgboost.XGBModel)` doesn't pick up the merlin models case because it doesn't inherit from `xgboost.XGBModel`.

oliverholworthy: > lightgbm merlin model case

We don't have LightGBM integration in Merlin Models currently. It would be a similar implementation to the XGBoost case though.

merlin/systems/dag/ops/fil.py
```diff
@@ -301,7 +301,7 @@ def export(
             **self.parameters,
         )
 
-        with open(node_export_path / "config.pbtxt", "w") as o:
+        with open(node_export_path / "config.pbtxt", "w", encoding="utf-8") as o:
```
oliverholworthy, line 304: pre-commit now checks encoding specified in open (probably since #150)

merlin/systems/dag/ops/fil.py
oliverholworthy: ✅  moved this line up to be alongside the other xgboost models in the elif as suggested

oliverholworthy (approved):

tests/unit/systems/fil/test_forest.py
```diff
@@ -96,6 +97,45 @@ def test_export(tmpdir):
     assert parsed_config.backend == "fil"
 
 
+def test_export_merlin_models(tmpdir):
```
oliverholworthy, line 100: This test only runs in Jenkins for now, can be activated in CPU / Github actions workflow by adding merlin models install in setup

PR #149, opened 2022-07-20
Load models from a model registry rather than path

nv-alaiacano:
This gives the option to load models into `PredictTensorflow` from any model registry, and includes an implementation for mlflow.

Example usage:

```python
PredictTensorflow.from_model_registry(
    MLFlowModelRegistry("my_production_model", "version-1", "http://tracking-uri")
)
```

The implementation simply gets the path to the model and uses it to call `PredictTensorflow`'s init method.

One issue that I ran into is that it seems like it would be nice to have this available for all implementations of `InferenceOperator`, but not all of those implementations accept a model path as an input - in fact only `PredictTensorflow` does. `FIL` only takes a model input as an input, and `TransformWorkflow` doesn't use a model at all.

I came up with a questionable solution of looking for `model_or_path` in the signature of the implementation class's `__init__` method (in fact, it should be the first argument to that method). I'd love to hear a suggestion of a better way to do this.

### DIFF
--- a/merlin/systems/dag/ops/operator.py
+++ b/merlin/systems/dag/ops/operator.py
@@ -4,9 +4,8 @@ import os
 import pathlib
 from abc import abstractclassmethod, abstractmethod
 from shutil import copyfile
-from typing import Optional
 
-import requests
+from merlin.systems.model_registry import ModelRegistry
 
 # this needs to be before any modules that import protobuf
 os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
@@ -125,9 +124,7 @@ class InferenceOperator(BaseOperator):
         return InferenceNode(selector)
 
     @classmethod
-    def from_mlflow_registry(
-        cls, name: str, version: str, mlflow_tracking_uri: Optional[str] = None, **kwargs
-    ) -> "InferenceOperator":
+    def from_model_registry(cls, registry: ModelRegistry, **kwargs) -> "InferenceOperator":
         """
         Loads the InferenceOperator from an MLflow Model Registry.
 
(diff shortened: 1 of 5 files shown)


merlin/systems/dag/ops/operator.py
```diff
@@ -120,6 +123,39 @@ def create_node(self, selector: ColumnSelector) -> InferenceNode:
...
+
+        Returns
+        -------
+        InferenceOperator
+            New node for Ensemble graph.
+        """
+
+        # validate that we have a model_or_path argument. This is a total hack and perhaps a smell
+        # that all inference operators should accept a model_or_path, or at least we introduce
+        # another layer of base class that does.
```
karlhigley, line 150: One trick I use sometimes when there are two options is to make one of them the type expected by the constructor and then add a class method that takes the other type and converts it. In this case, I'd probably pass the model type into constructors, and add a class method like `from_path` that loads the models and passes them to the constructors.

merlin/systems/model_registry.py
```diff
@@ -0,0 +1,61 @@
+from abc import abstractmethod
+
+import requests
+
+
+class ModelRegistry:
```
karlhigley, line 6: Since this only has an abstract method, maybe the class should be abstract too?

merlin/systems/dag/ops/operator.py
nv-alaiacano: That's a great idea. In the case of `TransformWorkflow` the `from_path` method could throw some kind of exception since it doesn't really make sense for that operator.

karlhigley: I think it would make sense if you want to load a Workflow from a file, which is not an uncommon thing to do.

karlhigley (approved):
🚀

PR #152, opened 2022-07-29
Clean up model input signature handling for tuples

karlhigley:
Partial fix for https://github.com/NVIDIA-Merlin/Merlin/issues/361#issuecomment-1191889455

### DIFF
--- a/merlin/systems/dag/ops/operator.py
+++ b/merlin/systems/dag/ops/operator.py
@@ -316,6 +316,9 @@ def _add_model_param(params, paramclass, col_schema, dims=None):
             )
         )
     else:
+        if col_schema.is_list:
+            value_count = col_schema.properties.get("value_count", {})
+            dims = [-1, value_count.get("max", 1)]
         params.append(
             paramclass(name=col_schema.name, data_type=_convert_dtype(col_schema.dtype), dims=dims)
         )
--- a/merlin/systems/dag/ops/operator.py
+++ b/merlin/systems/dag/ops/operator.py
@@ -1,7 +1,7 @@
 import json
 import os
 import pathlib
-from abc import abstractclassmethod, abstractmethod
+from abc import abstractmethod
 from shutil import copyfile
 
 from merlin.systems.model_registry import ModelRegistry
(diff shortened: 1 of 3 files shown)


merlin/systems/dag/ops/operator.py
```diff
@@ -297,3 +298,27 @@ def _schema_to_dict(schema: Schema) -> dict:
         }
 
     return schema_dict
+
+
+def _add_model_param(params, paramclass, col_schema, dims=None):
+    dims = dims if dims is not None else [-1, 1]
+    if col_schema.is_list and col_schema.is_ragged:
```
nv-alaiacano, line 305: Could we update `test_tf_op_exports_own_config` to give it a model with a tupled input as well so this new logic gets covered?

nv-alaiacano (approved):
lgtm aside from the one useful regression test.

I assume this `_add_model_param` function is living in `operator.py` instead of `tensorflow.py` because we expect to use it for other model frameworks as well (besides FIL), is that right?

merlin/systems/dag/ops/tensorflow.py
```diff
@@ -143,38 +116,69 @@ def _export_model(self, model, name, output_path, version=1):
...
         )
 
+        config.parameters["TF_GRAPH_TAG"].string_value = "serve"
+        config.parameters["TF_SIGNATURE_DEF"].string_value = "serving_default"
+
+        for _, col_schema in self.input_schema.column_schemas.items():
+            _add_model_param(config.input, model_config.ModelInput, col_schema)
+
+        for _, col_schema in self.output_schema.column_schemas.items():
+            _add_model_param(config.output, model_config.ModelOutput, col_schema)
```
nv-alaiacano, line 130: This logic would likely also need to be used in `PipelineableInferenceOperator.export` https://github.com/NVIDIA-Merlin/systems/blob/572dbf154015924106718d21f1896037d8a3bab6/merlin/systems/dag/ops/operator.py#L259-L274

merlin/systems/dag/ops/operator.py
karlhigley: Added a new test to cover the schema inference from the model file

PR #157, opened 2022-08-01
add xgboost serving example

radekosmulski:
This is a draft of the xgboost serving examples. One other thing I need to do is add an integration test

### DIFF
--- a/examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
+++ b/examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
@@ -82,7 +82,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "id": "0a2d3208",
    "metadata": {},
    "outputs": [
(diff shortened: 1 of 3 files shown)


bschifferer:
I know it is WIP, but I had some time and wanted to provide some initial feedback. The example looks great! 

Some feedback - it seems like a long list, but a lot is just format/text and super detailed :)

* Can we add a header (between license and first text) as we have in any other example? It is the title of the notebook
* Can we have a similar structure in the top section (e.g. providing learning objectives)
* I think we can combine "Download", "Data preprocessing" and "Train an XGBoost model" into "Prerequisite: Preparing the data and Training XGBoost" - we can add the reference to the other notebook and say that this is based on that notebook, thats the reason we do not go into details. Can we add Tagging the Target Column to the NVTabular workflow? Althougth the column is tagged by the received workflow - as we provide another NVT workflow, I can see how user will apply NVTabular to their raw data and dont have a target column defined
* Can we shorten the variable names? They are really nicely descriptive, but using them into other functions get pretty long commands e.g. `inference_preprocessing_workflow ` and `inference_preprocessing_workflow.output_schema`, `inference_schema`, etc.
* `an operator we can leverage to executing our workflow during serving` -> an operator we can leverage to executing our NVTabular workflow during serving
* Starting the Triton Server - can we be consistent with the other examples? https://github.com/NVIDIA-Merlin/systems/blob/main/examples/Serving-Ranking-Models-With-Merlin-Systems.ipynb provide an explanation and the user needs to start it him/herself. (Note: that the container description is not uptodate in the example)

That is a great test in the end (comparing the two predictions). Does the unittest cover it? Comparing it? I am not sure, if testbook will check the assert and then throw an error for the unittest, if they are different

radekosmulski:
Thank you very much for the review, @bschifferer, appreciate it! 🙂 I implemented the feedback, please let me know if any other changes would be needed.

Yes, if that assert at the end of the notebook triggers the test will fail 🙂 I stumbled into the issue with serialization and that assert worked okay with the prose of the notebook so opted to go down that route.

In fact, running the test in any of our containers will lead to it failing, because in order for the predictions to be the same on triton as locally we need a fix that is on merlin models main but that still hasn't made its way into our containers. I am not really sure how our CI container gets updated, maybe that is something we could talk about when we meet in a few 🙂

bschifferer:
If we can update the text for
```
For the --model-repository argument, specify the same value as the export_path that you specified previously in the ensemble.export method.
```

bschifferer:
Can we update using the unittest with 

``` 
with run_triton_server(modelpath) as client:
    execute cells
```

radekosmulski:
rerun tests

radekosmulski:
rerun tests

examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
```diff
@@ -0,0 +1,527 @@
+{
```
bschifferer, line 1: <p>In the title, it should be Serving an XGBoost Model with Merlin Systems -&gt; XG*B*oost. I think An and With should be lowercase letters.</p><p> </p><p>deployt -&gt; deploy</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

bschifferer, line 1: <p>You renamed the objects in the code to shorten them. We need to update the names in the text.</p><p> </p><p>Triton Inference Server -&gt; the first letter of each word is upper case</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

bschifferer, line 1: <p>It the text <code>We do so in a subprocess in order to continue our exploration in this notebook.</code> still correct? I dont think we do this anymore? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/157/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

bschifferer:
What is the difference between:
```
inf_schema = train_workflow.input_schema.remove_col('rating')
```
and
```
inf_workflow = train_workflow.remove_inputs(['rating'])
```

It seems in each case, we remove the input column rating - is that correct? Can we replace
`inf_schema` with `inf_workflow.input_schema` ?

```
inf_workflow = train_workflow.remove_inputs(['rating'])
inf_schema = inf_workflow.input_schema
```
?

bschifferer:
Can we update the unittest to use synthetic data? For example:
https://github.com/NVIDIA-Merlin/models/blob/main/tests/unit/tf/examples/test_01_getting_started.py

We inject a cell in the beginning, which mocks the function to return a synethic generated dataset. Otherwise, every unittest/CI run will download the movielens dataset

examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
radekosmulski: Yes, absolutely! Not sure how I missed that, apologies. Removing it now

radekosmulski: Fixed! Will capitalize titles like you advise going forward

radekosmulski: Good point -- changes made

radekosmulski:
> What is the difference between:
> 
> ```
> inf_schema = train_workflow.input_schema.remove_col('rating')
> ```
> 
> and
> 
> ```
> inf_workflow = train_workflow.remove_inputs(['rating'])
> ```
> 
> It seems in each case, we remove the input column rating - is that correct? Can we replace `inf_schema` with `inf_workflow.input_schema` ?
> 
> ```
> inf_workflow = train_workflow.remove_inputs(['rating'])
> inf_schema = inf_workflow.input_schema
> ```
> 
> ?

Yes, you are absolutely correct. I missed that this could be refactored like you mentioned, made the change

radekosmulski:
> > What is the difference between:
> > ```
> > inf_schema = train_workflow.input_schema.remove_col('rating')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> > inf_workflow = train_workflow.remove_inputs(['rating'])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems in each case, we remove the input column rating - is that correct? Can we replace `inf_schema` with `inf_workflow.input_schema` ?
> > ```
> > inf_workflow = train_workflow.remove_inputs(['rating'])
> > inf_schema = inf_workflow.input_schema
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> I suggest to use `.without()` method as well.

Thank you @rnyak for the suggestions! I switched to using `.without()` -- it is much nicer as it accepts a list, thank you! 🙂

bschifferer (approved):
It looks great!

PR #158, opened 2022-08-02
Pass model kwargs through to operator from_config methods

oliverholworthy:
Passing model kwargs (model_repository, model_name, model_version) through to operator `from_config` methods when used in an `OperatorRunner`.

## Motivation

This enables operators to use the location of the model repository to load artifacts that may be required for the operator to run. For example, a serialized model saved alongside the model conifg (or in the model version directory).  

Intended to be used by the Implict operator in #134 . We could also update the QueryFaiss operator to load the index from a relative path too instead of relying on a path in the model config. This would make the model repository self-contained.

### DIFF
--- a/merlin/systems/dag/op_runner.py
+++ b/merlin/systems/dag/op_runner.py
@@ -23,6 +23,7 @@ class OperatorRunner:
     def __init__(
         self,
         config,
+        *,
         model_repository="./",
         model_version=1,
         model_name=None,
--- a/merlin/systems/dag/op_runner.py
+++ b/merlin/systems/dag/op_runner.py
@@ -18,7 +18,16 @@ import json
 
 
 class OperatorRunner:
-    def __init__(self, config, repository="./", version=1, kind=""):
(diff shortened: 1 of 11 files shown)


merlin/systems/dag/op_runner.py
```diff
@@ -18,7 +18,17 @@
 
 
 class OperatorRunner:
-    def __init__(self, config, repository="./", version=1, kind=""):
```
oliverholworthy, line 21: removed `kind` since it doesn't appear to be used currently. Can be added back if required in future

merlin/systems/dag/op_runner.py
```diff
@@ -18,7 +18,17 @@
...
+    """Runner for collection of operators in one triton model."""
+
+    def __init__(
+        self,
+        config,
+        *,
```
oliverholworthy, line 26: this forces the following parameters to be passed as keyword arguments.

karlhigley (approved):

PR #159, opened 2022-08-02
(WIP) Experiment with combining `dtype`/`is_list`/`is_ragged` into dataclass

karlhigley:
These three properties in `ColumnSchema`s get passed around together a lot. I think it would help us clean up situations like this one where we want to build cross-framework dtype translation (e.g. `Merlin` dtypes <-> `Feast` dtypes, but also TF/Torch/cuDF/etc.)


merlin/systems/dag/ops/feast.py
```diff
@@ -9,13 +10,22 @@
...
-    ValueType.INT64: (np.int64, False, False),
-    ValueType.INT32: (np.int32, False, False),
-    ValueType.FLOAT: (np.float, False, False),
-    ValueType.INT64_LIST: (np.int64, True, True),
-    ValueType.INT32_LIST: (np.int32, True, True),
-    ValueType.FLOAT_LIST: (np.float, True, True),
+
+
+@dataclass
+class MerlinDtype:
```
nv-alaiacano, line 16: This should probably live somewhere more accessible than the feast.py module if it's something we'd like to reuse outside of this one file. It seems like something that could be in `core` eventually, but for now perhaps `merlin/systems/dtypes.py` or similar so we could adopt it in the systems repo?

karlhigley: Yeah, this was originally proposed as something that would go in Merlin Core but faced rather stiff resistance, so I wanted to try it out here first and demonstrate what it could look like

nv-alaiacano:
rerun tests

nv-alaiacano:
rerun tests

nv-alaiacano:
rerun tests

nv-alaiacano:
rerun tests

merlin/systems/dag/ops/feast.py
karlhigley: Another thing that should happen here while I'm thinking about it: rework the two boolean flags into an `Enum` with types `Scalar`, `FixedList`, and `RaggedList`

karlhigley: (Since there's no such thing as a ragged scalar field, using two booleans allows an invalid state)

karlhigley:
Closing since this change (if it happens) will end up in `core`

PR #160, opened 2022-08-05
Add unit tests for QueryFeast op

nv-alaiacano:
I was reading through some ops to understand the create/export lifecycle and decided to add some tests to make sure I get it well. 

I didn't add a test for `from_feature_view` because the required mocking of the feast components is pretty deep and would probably be better done as an integration test.

### DIFF
--- a/.github/workflows/cpu-ci.yml
+++ b/.github/workflows/cpu-ci.yml
@@ -35,6 +35,8 @@ jobs:
         python -m pip install tensorflow
         # required for ensemble tests
         python -m pip install xgboost
+        # required for feast unit tests
+        python -m pip install "feast<0.20"
     - name: Lint with flake8
       run: |
         flake8 .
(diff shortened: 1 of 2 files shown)


.github/workflows/cpu-ci.yml
```diff
@@ -35,6 +35,8 @@ jobs:
         python -m pip install tensorflow
         # required for ensemble tests
         python -m pip install xgboost
+        # required for feast unit tests
+        python -m pip install "feast<0.20"
```
karlhigley, line 39: I think we'd also have to add this to the stuff that gets installed on the Jenkins machine

nv-alaiacano: If I've pulled the thread correctly, it looks like it's already there: https://github.com/NVIDIA-Merlin/Merlin/blob/a1cc48fe23c4dfc627423168436f26ef7e028204/ci/dockerfile.ci#L18

karlhigley (approved):

PR #164, opened 2022-08-11
test to ensure samplingsoftmax works in an in ensemble

nv-alaiacano:
This is currently failing but I think that it is expected to work.

cc @rnyak

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -2,6 +2,7 @@
 -r requirements-dev.txt
 
 # This contains common libraries for testing.
+merlin-models>=0.6.0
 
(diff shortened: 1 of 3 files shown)


karlhigley:
Did this issue already get fixed or is the test in this PR not running? 🤔

nv-alaiacano:
The description is out of date - haven't gotten a chance to update it. The current status is:

This tests that the ensemble schema is correct when using SoftmaxSampling, and the tests pass. We are correctly passing the _schema information_ about the inputs through to the outputs.

I'm not able to actually run triton server to verify that the inference request works because I'm still running into some of last week's CI issues.

PR #165, opened 2022-08-12
Update `versioneer` from 0.21 to 0.23

oliverholworthy:
### Goals :soccer:

Restore editable install support with latest version of `setuptools` (without requiring setting any environment variables to activate legacy support).

### Implementation Details :construction:

Ran the following

```shell
pip install --upgrade versioneer
versioneer install
```

There are a few changes since we last configured versioneer (0.21).

The relevant change we're after to fix the editable install is in release 0.23 which added a patch for compatibility with the new  `setuptools` release.

### Testing Details :mag:

Checked that the following works locally:

```shell
pip install -e .
```


versioneer.py
```diff
@@ -1820,20 +1845,25 @@ def run(self):
...
-        from distutils.command.build_py import build_py as _build_py
+        from setuptools.command.build_py import build_py as _build_py
 
     class cmd_build_py(_build_py):
         def run(self):
             root = get_root()
             cfg = get_config_from_root(root)
             versions = get_versions()
             _build_py.run(self)
+            if getattr(self, "editable_mode", False):
```
oliverholworthy, line 1863: This is the part that handles the new `setuptools` version

karlhigley (approved):

PR #167, opened 2022-08-15
Add preliminary support for serving PyTorch models with TorchScript

karlhigley:
Support for serving with the Python back-end is planned, but not yet included here.


tests/unit/systems/torch/model_config
```diff
@@ -0,0 +1,46 @@
+#### Broken ####
```
oliverholworthy, line 1: Is this file for debug or relevant to the functionality here?

tests/unit/systems/torch/test_torch.py
```diff
@@ -0,0 +1,197 @@
...
+
+class CustomModel(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.linear = torch.nn.Linear(3, 1)
+        # self.flatten = torch.nn.Flatten(0, 1)
+
+    def forward(self, input_dict: Dict[str, Tensor]):
+        linear_out = self.linear(input_dict["input"].to(self.linear.weight.device))
+        # flatten_out = self.flatten(linear_out.to(self.linear.weight.device))
```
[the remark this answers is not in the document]
oliverholworthy: (for when we follow-up with the work on the pytorch python-backend)?

merlin/systems/dag/ops/pytorch.py
```diff
@@ -0,0 +1,191 @@
...
+                    dims = [-1, value_count["max"]]
+
+            config.input.append(
+                model_config.ModelInput(
+                    name=col_name, data_type=_convert_dtype(col_schema.dtype), dims=dims
+                )
+            )
+
+        for col_name, col_schema in self.output_schema.column_schemas.items():
+            # this assumes the list columns are 1D tensors both for cats and conts
```
oliverholworthy, line 159: This comment is non-blocking for this PR. I'm wondering if there are any ways to inspect the output shape you considered but didn't implement. (e.g. running a forward pass through the model with synthetic data?) to allow us to remove this restriction. As we have discussed before, maybe only a priority if this assumption is something that is easy to break with Merlin Models.

oliverholworthy: I suppose related to this TODO comment
https://github.com/NVIDIA-Merlin/systems/blob/905143cb0a695abc7f94f9c69bb81fd31f46b802/merlin/systems/dag/ops/pytorch.py#L66

karlhigley: Yeah, there are definitely issues outstanding here; I would not recommend trying to use this for anything real yet. Mostly want to get the preliminary support merged so that it's easier to test out with the Torchscript versions of Transformers4Rec models. I expect to find a lot of further issues when we do that.

karlhigley:
@oliverholworthy Do these test failures make any sense to you? I'm also seeing them on the `main` branch after merging the Implicit PR.

oliverholworthy:
@karlhigley The tests are passing locally for me and did pass before the Implicit PR was merged.

Every op should have `operator_names` in the config. So the fact that this KeyError is cropping up suggests to me that there is a triton server that didn't shutdown with different config (perhaps from the NVTabular tests?). Perhaps the triton test isolation issue again?

karlhigley:
If I pull `merlin_ci_runner:latest`, `pip install implicit` in the container, and run the tests from the latest commit of this PR, it seems like a repeatable failure

oliverholworthy:
Hmm, I've pulled the latest image now and tried again, the tests still pass for me. Not sure what could be different about my environment for that to be the case. 

Perhaps we can disable that test for now while we figure it out if it's blocking PRs now.

oliverholworthy (approved):

PR #168, opened 2022-08-15
Build the documents with tox

mikemckiernan:


tox.ini
```diff
@@ -39,7 +39,16 @@ commands =
...
 ; Generates documentation with sphinx. There are other steps in the Github Actions workflow
 ; to publish the documentation on release.
-changedir = {toxinidir}/docs
+changedir = {toxinidir}
 deps = -rrequirements-docs.txt
 commands =
-    python -m sphinx.cmd.build -P -b html source build
+    python -m sphinx.cmd.build -P -b html docs/source docs/build/html
+
+[testenv:docs-multi]
```
mikemckiernan, line 47: @nv-alaiacano , is there a way to avoid creating a second environment? Or a way to run the python command in one case and sphinx-multiversion in another case--like an env var or something?

mikemckiernan:
API seems to show up: https://nvidia-merlin.github.io/systems/review/pr-168/generated/merlin.systems.dag.ops.workflow.TransformWorkflow.html#merlin.systems.dag.ops.workflow.TransformWorkflow

tox.ini
nv-alaiacano: hm i'm not sure. it would be nice if we could cache the `.tox` directory to avoid rebuilding every time. I'll look into that.

karlhigley (approved):

PR #170, opened 2022-08-16
Organize the `tests` directory more clearly

karlhigley:


oliverholworthy (approved):
This is certainly clearer than before. Are we aiming for a 1:1 mirror of the main source directory in the test structure?

nv-alaiacano (commented):
I really like the rearrangement of the actual tests - well done for digging through all of this!

I can see that the directory structure of the tests is by framework (feast, fil, nvtabular, etc), which I like, but maybe we can be consistent with the file names inside of those directories with a format such as `test/unit/systems/[framework]/test_[function].py`. For example:

```
test/unit/systems/feast/test_op.py
test/unit/systems/feast/test_ensemble.py  <- if it existed

test/unit/systems/tf/test_op.py   <- if it existed
test/unit/systems/tf/test_ensemble.py
```

This would set us up to have a good template for what kinds of tests should exist for each type of op. For example we'd want to make sure exports properly, and that the actual `transform` method does what it should. A separate `test_ensemble.py` for each op would make sure that it works as expected in an ensemble (passes schema correctly, etc). And if there is other framework-specific functionality, then we could have `test_[functionality].py` for that.

karlhigley:
Yeah, I like that proposal for consistent naming of the test files, I'll add that here.

I think 1:1 correspondence with at least the directory structure of the source files makes sense, but I don't think they're currently very well organized either. Propose to re-arrange them in a separate PR, since that will affect imports and we may want to deal with making the imports backwards compatible.

PR #176, opened 2022-08-18
Write an integration test for T4R Torchscript serving

karlhigley:
Depends on https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/486


merlin/systems/dag/ensemble.py
```diff
@@ -72,18 +73,20 @@ def export(self, export_path, version=1):
...
-            ensemble_config.input.append(
-                model_config.ModelInput(
-                    name=col_name, data_type=_convert_dtype(col_schema.dtype), dims=[-1, -1]
-                )
+        for _, col_schema in self.graph.input_schema.column_schemas.items():
+            add_model_param(
+                ensemble_config.input,
+                model_config.ModelInput,
+                col_schema,
+                col_schema.properties.get("shape", None) or compute_dims(col_schema),
```
karlhigley, line 81: I'm using an ad-hoc property here since we don't yet have a `.shape` attribute on `ColumnSchema`, but I'd expect to switch over to `col_schema.shape` once that's available.

merlin/systems/dag/ensemble.py
```diff
@@ -116,17 +119,34 @@ def export(self, export_path, version=1):
...
-                for input_col_name in node.input_schema.column_names:
+                for input_col_name, input_col_schema in node.input_schema.column_schemas.items():
                     source = _find_column_source(node.parents_with_dependencies, input_col_name)
                     source_id = node_id_lookup.get(source, None)
                     in_suffix = f"_{source_id}" if source_id is not None else ""
-                    config_step.input_map[input_col_name] = input_col_name + in_suffix
-
-                for output_col_name in node.output_schema.column_names:
+                    if input_col_schema.is_list and input_col_schema.is_ragged:
+                        config_step.input_map[input_col_name + "__values"] = (
```
karlhigley, line 127: I'd really like to create some abstraction to represent both of the arrays that represent ragged multi-hot columns, but we're not there yet so we have to handle this case explicitly in several places.

merlin/systems/dag/ops/operator.py
```diff
@@ -306,8 +310,7 @@ def _schema_to_dict(schema: Schema) -> dict:
     return schema_dict
 
 
-def _add_model_param(params, paramclass, col_schema, dims=None):
-    dims = dims if dims is not None else [-1, 1]
+def add_model_param(params, paramclass, col_schema, dims=None):
```
karlhigley, line 313: Computing the dims was a second responsibility of this function, so it moved to a separate utility function (above.)

merlin/systems/dag/ops/softmax_sampling.py
```diff
@@ -89,7 +89,7 @@ def compute_output_schema(
         self, input_schema: Schema, col_selector: ColumnSelector, prev_output_schema: Schema = None
     ) -> Schema:
         """Describe the operator's outputs"""
-        return Schema([ColumnSchema("ordered_ids", dtype=np.int32, is_list=True, is_ragged=True)])
+        return Schema([ColumnSchema("ordered_ids", dtype=np.int32, is_list=True, is_ragged=False)])
```
karlhigley, line 92: The value of `is_ragged` didn't used to matter here, since we didn't have a clear differentiation between how ragged and non-ragged lists were handled. It does now though, and this is a fixed-length list since we're picking the top-k elements.

merlin/systems/triton/__init__.py
```diff
@@ -34,42 +34,103 @@
 )
 
 
-def convert_df_to_triton_input(column_names, batch, input_class=grpcclient.InferInput):
-    columns = [(col, batch[col]) for col in column_names]
-    inputs = []
-    for i, (name, col) in enumerate(columns):
-        if is_list_dtype(col):
+def convert_df_to_triton_input(schema, batch, input_class=grpcclient.InferInput, dtype="int32"):
```
karlhigley, line 37: We need the schema here to get the column shapes, but it's handy in a few other ways too (like checking `is_ragged`.)

tests/unit/examples/test_serving_ranking_models_with_merlin_systems.py
```diff
@@ -105,5 +106,12 @@ def test_example_04_exporting_ranking_models(tb):
     configure_tensorflow()
     from merlin.systems.triton.utils import run_ensemble_on_tritonserver
 
-    response = run_ensemble_on_tritonserver("/tmp/data/ensemble/", outputs, batch, "ensemble_model")
+    # The schema contains int64s, while the actual data contains int32s. Not sure why.
+    schema = Schema(
+        [col_schema.with_dtype(np.int32) for col_schema in schema.column_schemas.values()]
+    )
```
karlhigley, line 112: I think what's going on here is that the schema never matched the data, but the conversion from dataframe to Triton input didn't used to coerce the dtypes to schema dtypes.

tests/unit/systems/utils/triton.py
```diff
@@ -1,40 +0,0 @@
...
-from tritonclient import grpc as grpcclient
-
-triton = pytest.importorskip("merlin.systems.triton")
-data_conversions = pytest.importorskip("merlin.systems.triton.conversions")
-
-TRITON_SERVER_PATH = find_executable("tritonserver")
-from merlin.systems.triton.utils import run_triton_server  # noqa
-
-
-def _run_ensemble_on_tritonserver(
```
karlhigley, line 28: We've had two versions of this method kicking around for a while, and the public version is more convenient to use since it automatically converts the response to a dictionary.

karlhigley (commented):
Since PyTorch/Torchscript handles array dimensions slightly differently than Tensorflow, this ended up requiring deeper changes than I expected.

examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
```diff
@@ -430,7 +430,7 @@
     "import tritonclient.grpc as grpcclient\n",
     "\n",
     "ten_examples = train.compute().drop(columns=['rating', 'title', 'rating_binary'])[:10]\n",
-    "inputs = convert_df_to_triton_input(inf_workflow.input_schema.column_names, ten_examples, grpcclient.InferInput)\n",
```
karlhigley, line 433: The change to the signature of `convert_df_to_triton_input` will also affect the downstream notebooks in the Merlin repo.

karlhigley:
rerun tests

merlin/systems/dag/ops/operator.py
jperez999: I like this fall back.

jperez999 (approved):

examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
rnyak: is the change only replacing this line with `inputs = convert_df_to_triton_input(workflow.input_schema, batch, grpcclient.InferInput)\n`?

karlhigley:
rerun tests

PR #179, opened 2022-08-19
add test for QueryFeast.from_feature_view

nv-alaiacano:
Adds a regression test for the bug fix in https://github.com/NVIDIA-Merlin/systems/pull/178

I also caught an error with how the  `assert_called_with` method was used on the `QueryFeast` mock in another test and fixed it.

### DIFF
--- a/tests/unit/systems/ops/feast/test_op.py
+++ b/tests/unit/systems/ops/feast/test_op.py
@@ -1,4 +1,5 @@
 import json
+from datetime import timedelta
 from unittest.mock import MagicMock, patch
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


tests/unit/systems/ops/feast/test_op.py
```diff
@@ -43,16 +44,103 @@ def test_feast_config_round_trip(tmpdir):
...
+                ColumnSchema(name="prefix_float_feature", dtype=np.float32),
+                ColumnSchema(
+                    name="prefix_int_list_feature_1", dtype=np.int32, is_list=True, is_ragged=True
+                ),
+                ColumnSchema(name="prefix_int_list_feature_2", dtype=np.int32, is_list=True),
+                ColumnSchema(
+                    name="prefix_float_list_feature_1",
+                    dtype=np.float32,
+                    is_list=True,
+                    is_ragged=True,
```
nv-alaiacano, line 114: I don't quite understand why `_1` is ragged but `_2` is not ragged for these multi-hot columns.

karlhigley: IIRC, because `_1` has a variable length and `_2` does not, but this really should be a single ragged column in the schemas.

karlhigley (approved):

nv-alaiacano:
rerun tests

PR #183, opened 2022-08-24
add testenv to run merlin tests

nv-alaiacano:
This will

* Clone the NVIDIA-Merlin/Merlin repo to the directory `Merlin-$COMMIT_SHA` for this branch's commit sha.
* Run `pip install .` for this branch of systems (necessary??)
* Run the Merlin repo tests

Right now now, I expect the cloned repo to be cleaned up by Jenkins. Tox doesn't really have a "cleanup" phase, unfortunately.

I am also manually updating `dask` and `distributed` to  >=2022.3.0. For some reason, our jenkins image is somehow downgrading the version which causes an error when running the Merlin notebook tests. Work is in progress to sort that out, but we are able to unblock this PR by upgrading it here.

Update:

* There is a flakiness issue with the Merlin integration tests, so I downgraded this to only run unit tests for now.
* I'd like to split this into its own build check so that it can run on jenkins in parallel with the systems repo tests. TBD.

### DIFF
--- a/requirements-gpu.txt
+++ b/requirements-gpu.txt
@@ -2,3 +2,5 @@ feast<=0.20.0
 faiss-gpu
 pytest
 pytest-cov
+dask>=2022.3.0
+distributed>=2022.3.0
--- a/requirements-gpu.txt
+++ b/requirements-gpu.txt
@@ -1,8 +1,4 @@
--r requirements.txt
 feast<=0.20.0
 faiss-gpu
 pytest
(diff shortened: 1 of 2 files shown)


nv-alaiacano:
rerun tests

nv-alaiacano:
rerun tests

nv-alaiacano:
rerun tests

karlhigley (approved):

PR #184, opened 2022-08-26
Specify `is_ragged=False` explicitly in the Feast tests

karlhigley:
The second array representing a ragged column should have `is_ragged=False` since the number of items is fixed.


nv-alaiacano (approved):

karlhigley:
rerun tests

PR #186, opened 2022-08-29
Remove the `tritonclient` importorskip from `conftest.py`

karlhigley:
This causes all tests to be skipped in environments without Triton, but we do have tests that should run fine on CPU without Triton.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -50,10 +50,6 @@ except ImportError:
 
 
 REPO_ROOT = Path(__file__).parent.parent
-
-grpcclient = pytest.importorskip("tritonclient.grpc")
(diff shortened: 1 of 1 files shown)


karlhigley:
I'm going to add `tritonclient` to the test dependencies instead

oliverholworthy:
Once #187 is merged. We could then follow-up with a PR to remove all the tritonclient `importorskip` lines from the tests, switching to a regular import. It seems like this PR may still be relevant? Unless we want to maintain the ability to run the tests in a test environment where tritonclient isn't installed?

karlhigley:
Yeah, that's a good thought. I'll put together a PR like that.

PR #189, opened 2022-08-30
Add extra requirements files to `setup.py` as optional installs

karlhigley:
Hoping this will make it a bit easier to install downstream libraries and their test requirements from the `tox` envs for `core`.


karlhigley:
rerun tests

PR #194, opened 2022-09-05
add implicit example

radekosmulski:
resolves NVIDIA-Merlin/models#577


bschifferer:
@karlhigley do you know which import skip we need for this unittest?

```
pytest.importorskip("merlin.models")
```

do we need to add implicit?

examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb
```diff
@@ -0,0 +1,451 @@
+{
```
bschifferer, line 1: <p>The logo is the xgboost-model-with-merlin-systems logo</p><p> </p><p><code>We train an XGBoost model to predict the rating based on user and item features and proceed to deploy it to the Triton Inference Server.</code>   it is XGBoost here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/194/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

bschifferer, line 1: <p>Title contains XGBoost</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/194/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

bschifferer, line 1: <p>Not sure, if we can set <code>use_gpu=True</code> for the different unittest environments. Let's clarify </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/systems/pull/194/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

bschifferer:
@karlhigley @benfred The example uses `BayesianPersonalizedRanking`. How do we need to setup the example for the unittest and different environments? Currently, we use
```
model = BayesianPersonalizedRanking(use_gpu=True)
```

Would that break some unit tests, which runs on CPU? Does the default behavior switch between cpu/gpu?

examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb
radekosmulski: fixed

radekosmulski: changed

karlhigley:
`importorskip("implicit")` sounds right to me. I would suspect that `(use_gpu=True)` wouldn't work on CPU CI, but the tests seem to be passing so... 🤔

examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb
radekosmulski: discussed on slack, implemented the suggested changes, our test now should be compatible with the current state of the CI, but should also execute both on CPU and GPU when triton is available on the machine (which is not true currently for CPU only systems)

bschifferer (approved):

PR #201, opened 2022-09-08
Mark the version test so it can be skipped in other repos

karlhigley:
When we install from source in the CPU CI tests in other repos, the version numbers won't pass this test. Marking it so it can be skipped with `pytest -m not version`.


nv-alaiacano (approved):

karlhigley:
rerun tests

PR #204, opened 2022-09-15
Rework Systems operators to use `DictArray` and `LocalExecutor` from Merlin Core

karlhigley:
Depends on NVIDIA-Merlin/core#139 and NVIDIA-Merlin/core#146

We've had the concept of an `InferenceDataFrame` for a while, but it's really just a wrapper around a dictionary of arrays. That data structure is useful in a bunch of places, so this swaps out `InferenceDataFrame` for a similar class from Merlin Core called `DictArray`.


merlin/systems/dag/op_runner.py
```diff
@@ -50,8 +52,14 @@ def __init__(
 
     def execute(self, tensors):
         """Run transform on multiple operators"""
+
+        selector = ColumnSelector("*")
```
karlhigley, line 56: Systems operators have not historically paid very close attention to column selectors, and have defaulted to processing all input columns (except in a few particular cases like `UnrollFeatures`.) We've tried to align the Systems operator API more closely with the NVTabular operator API so that we can use `LocalExecutor`, but that API change also affects older code like this class that's used with Triton ensembles. Here we just supply a default selector that selects everything, which mirrors the existing behavior with the new API.

merlin/systems/dag/ops/faiss.py
```diff
@@ -131,39 +142,48 @@ def export(
...
+        full_path.mkdir(parents=True, exist_ok=True)
         copy2(self.index_path, new_index_path)
         self.index_path = str(new_index_path)
-        return super().export(path, input_schema, output_schema, self_params, node_id, version)
 
-    def transform(self, df: InferenceDataFrame) -> InferenceDataFrame:
+        if backend == "ensemble":
+            return super().export(path, input_schema, output_schema, self_params, node_id, version)
+        else:
+            return ({}, [])
```
karlhigley, line 160: When we run the FAISS operator in `LocalExecutor` instead of in a Triton ensemble, we still need to execute the `export` method in order to write out the FAISS index file to disk, but we don't need a Triton config for the operator since it no longer has its own directory in the Triton model repository. To make that work with the method signature, we return a null object here.

merlin/systems/dag/ops/feast.py
```diff
@@ -185,6 +186,13 @@ def __init__(
         self.store = FeatureStore(repo_path=repo_path)
         super().__init__()
 
+    def __getstate__(self):
+        # feature store objects aren't picklable - exclude from saved representation
```
karlhigley, line 190: We don't want to try to pickle anything that either doesn't play nicely with pickle or that we could just as easily save to disk, so we exclude external files that can be loaded from disk by the ops.

requirements/dev.txt
```diff
@@ -3,7 +3,6 @@ tritonclient[all]
 
 # linting
 pylint==2.7.4
-click==8.0.3
```
karlhigley, line 6: This was in the requirements twice. Oops!

tox.ini
```diff
@@ -16,6 +16,8 @@ commands =
 deps = -rrequirements/test-cpu.txt
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git
+    python -m pip install .
```
karlhigley, line 19: Turns out installing the version-under-test is important, since Triton uses the version that's installed in the tests.

oliverholworthy: we don't currently test with Triton in the GitHub actions tests. So I assume this comment was intended to refer to the change below on the `test-gpu` env. No harm running the install here too, but I suppose not strictly required.

I recall that this wasn't possible before due to the way we were running the tests in Jenkins. To check my understanding: was this possibility of making Triton pick up the version-under-test unlocked by the transition to tox because it's now running in a virtual env that is isolated from the other tests being run?

karlhigley: I think you have it exactly right. The installed version shouldn't matter in CPU, but on GPU using tox lets us install the version in the current PR without stomping on the version installed in the CI container that's used by the tests in other repos (currently just `Merlin`.)

merlin/systems/triton/models/executor_model.py
```diff
@@ -0,0 +1,145 @@
...
+                responses.append(
+                    pb_utils.InferenceResponse(
+                        tensors=[], error=f"{exc_type}, {exc_value}, {tb_string}"
+                    )
+                )
+
+        return responses
+
+
+def _parse_model_repository(model_repository: str) -> str:
```
oliverholworthy, line 134: Looks like this was intended to be used in `TritonPythonModel`?

karlhigley: Good catch!

merlin/systems/dag/ops/fil.py
```diff
@@ -90,7 +87,20 @@ def compute_input_schema(
...
+
+    def export(
+        self,
+        path: str,
+        input_schema: Schema,
+        output_schema: Schema,
+        params: dict = None,
+        node_id: int = None,
+        version: int = 1,
+        backend: str = "ensemble",
```
oliverholworthy, line 102: This `backend` doesn't appear to be used in this method. I'm guessing it's here for consistency of export signatures?

oliverholworthy: However, There is one thing confusing me about this `backend` parameter. It's that the value of this `backend` for the `PipelineableInferenceOperator` export method means something different. In that method the backend  corresponds to the backend that will be written out in the Triton config. Which suggest that the `ensemble` is only valid for the `Ensemble` and not for all operators.

merlin/systems/dag/ops/faiss.py
```diff
@@ -131,39 +142,48 @@ def export(
         self_params.update(params)
         index_filename = os.path.basename(os.path.realpath(self.index_path))
 
-        # set index path to new path after export
-        full_path = pathlib.Path(path) / f"{node_id}_{QueryFaiss.__name__.lower()}" / str(version)
+        if backend == "ensemble":
```
oliverholworthy, line 145: Why does this conditional export only apply to the Faiss op and not all the others?

merlin/systems/dag/ops/fil.py
karlhigley: > the backend corresponds to the backend that will be written out in the Triton config

This is always true, but there's a wobble in the logic here. An operator that isn't exportable for a given format (e.g. `ensemble`) won't have its `export` method called, but some operators need to write artifacts to disk in cases where they won't write out a Triton model config. In those cases, we currently consider them exportable for the `ensemble` back-end, but return an empty model config from the `export` method.

We could have resolved these weird inconsistencies by introducing back-ends/run-times/adapters in this PR (since then each back-end could have its own implementation of `export` for any given operator via the operator implementation table), but we wrote it this way in part to motivate the refactor in that direction.

merlin/systems/dag/ops/faiss.py
karlhigley: `QueryFAISS` is the only Python op currently saving additional files to disk (aside from the Triton config and `model.py`.) In an `ensemble` export, we write the FAISS index file into the `n_queryfaiss` directory in the model repository, but that directory doesn't exist with an `executor` export, so instead we write it to the `executor_model` directory. If other ops were also writing additional files, this conditional would apply there too.

(But should disappear when we introduce backends/runtimes/adapters, since then we can have separate `export` methods for each export format.)

oliverholworthy: The Implicit op also writes out files to it's directory. Is that a different kind of file in this context?

merlin/systems/dag/ops/fil.py
oliverholworthy: In the context of this `PredictForest` operator, which currently has `exportable_backends = ["ensemble", "executor"]`. Based on what you described I would assume that for operators that include `executor` in `exportable_backends` then we would need to perform a conditional logic in the `export` method.

oliverholworthy: If I'm understanding this correctly. I think what this means is that we'll end up having unused models written out to the model repository (if the op contains `executor` in `exportable_backends`, then we'll be calling the transform method  as part of the LocalExecutor), but we'll also be exporting the node in the same way as for the ensemble case (if there's no conditional logic similar to the Faiss operator). I suppose that doesn't break any functionality, but maybe not the intention?

karlhigley: > for operators that include executor in exportable_backends then we would need to perform a conditional logic in the export method
It depends on how the operator is structured. For operators that run in custom Triton back-ends, we still need to export a separate model directory. For operators that wrap one operator in another (e.g. in order to align the schema of the outer operator with the graph), the outer wrapper op runs in the `Executor` (in plain Python) but the inner op runs in a custom back-end.

karlhigley: > I suppose that doesn't break any functionality (since, if the op contains executor in exportable_backends, then we'll be calling the transform method as part of the LocalExecutor), but maybe not the intention?

I think this is a case where we're using the operator's ensemble export twice:
* Once in the actual export of the full ensemble to run as a Triton ensemble
* Once for exporting the single operator as a separate model directory to be used in `Executor` mode (still required to use specific non-Python Triton back-ends)

merlin/systems/dag/ops/faiss.py
karlhigley: Ah, good point! That should probably write its files to the `executor_model` directory too then.

karlhigley:
Tests should pass once we merge https://github.com/NVIDIA-Merlin/core/pull/153 (small fixes to handle list dtypes)

oliverholworthy:
> | 1_predictimplicit   | 1       | UNAVAILABLE: Internal: ImportError: cannot import name 'InferenceDataFrame' from 'merlin.systems.dag.ops.operator' (/var/jenkins_home/workspace/merlin_systems/systems/.tox/test-gpu/lib/python3.8/site-packages/merlin/systems/dag/ops/operator.py) |

I think the reason we're getting this error may be due to the example notebook tests not writing the data out to a tmpdir. So it's using a model repository directory from a previous run (which still had the `InferenceDataFrame`).

We might need to update that to write out to a temporary directory

oliverholworthy (approved):
Let's merge once the tests are passing.  #214 fixes the issue with the examples

karlhigley:
No idea what's going on with the Implicit memory allocation failure here 😓

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

PR #212, opened 2022-10-07
Create a separate directory parallel to `unit` for version tests

karlhigley:
This test fails all the time in development and other repos, since dev versions aren't `>=0.6.0`.


nv-alaiacano (approved):

oliverholworthy:
Is the idea that this test will run prior to release but after we've made the next tag? And if that isn't happening already do we need to change something about the github actions workflow?

karlhigley:
Yes, something needs to change somewhere in the release process, but I'm not sure where.

PR #223, opened 2022-11-08
Introduce and apply a basic version of the `Column` abstraction

karlhigley:
Depends on https://github.com/NVIDIA-Merlin/core/pull/163

This PR takes some early steps to integrate a single array version of a `Column` class inside `DictArray` and get the Systems tests back to passing. Since this is a partial version of `Column`, we're not yet reaping the benefits as far as cleaning up the code goes and there's a lot that's still gross in this version of the code. We think it's at least "not worse" though, so in theory this small PR might be merge-able. There's _a lot_ of further refactoring to do, so we'll have to figure out what size we want the PRs to be.

### DIFF
--- a/merlin/systems/triton/models/oprunner_model.py
+++ b/merlin/systems/triton/models/oprunner_model.py
@@ -29,17 +29,13 @@ import pathlib
 import sys
 import traceback
 
-import numpy as np
 import triton_python_backend_utils as pb_utils
 
-from merlin.core.dispatch import HAS_GPU
-from merlin.systems.dag import DictArray
 from merlin.systems.dag.op_runner import OperatorRunner
-
-if HAS_GPU:
-    import cupy
-else:
-    cupy = None
+from merlin.systems.triton.conversions import (
+    dict_array_to_triton_response,
+    triton_request_to_dict_array,
+)
 
 
 class TritonPythonModel:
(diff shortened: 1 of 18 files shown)


karlhigley:
rerun tests

merlin/systems/triton/models/oprunner_model.py
```diff
@@ -110,7 +117,10 @@ def execute(self, requests):
                         output_tensors.append(data)
                         continue
                     data = data.get() if hasattr(data, "get") else data
-                    tensor = pb_utils.Tensor(name, data)
+                    if not isinstance(data.values, np.ndarray):
+                        tensor = pb_utils.Tensor(name, cupy.asnumpy(data.values))
```
oliverholworthy, line 121: What is the connection with DictArray/Column and this `cupy/numpy` conversion (and in the FIL op too)

karlhigley: Previously we allowed any array type in `DictArray`, but since it's hard to build on top of the varying APIs of the different array/tensor libraries, we're heading in a direction where the only allowed types will be Numpy and CuPy arrays (which gives us a consistent API to build on.) That means for other types, like Triton tensors, we'll need to convert them across to Numpy or CuPy, hopefully via zero-copy transfers. Unfortunately, there's not currently a way to do that with Triton GPU tensors and CuPy arrays because their dlpack implementations are not compatible with each other. So we work around that here by forcing the data to be moved from GPU to CPU. It's a little gross, but it's the best way we've been able to figure out how to do it without using another library (e.g. PyTorch) as an intermediary.

oliverholworthy (approved):

PR #224, opened 2022-11-10
Add multi-array `Columns` to `DictArray`

karlhigley:


karlhigley:
rerun tests

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,149 @@
...
         super().__init__()
 
         if isinstance(values, Column):
             raise TypeError("doubly nested columns")
 
         self.values = values
+        self.row_lengths = row_lengths
         self.dtype = values.dtype
 
+        if isinstance(values, np.ndarray):
```
karlhigley, line 54: We hadn't intended to tackle the scope of moving data between host and device memory here, but it cropped up in some of the Feast tests so we went ahead and did it.

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,149 @@
...
+    def _device_move(self, fn):
+        if not cupy:
+            raise ValueError("Unable to move data across devices without a GPU")
+        self.values = fn(self.values)
+        if self.row_lengths:
+            self.row_lengths = fn(self.row_lengths)
+
     def __getitem__(self, index):
-        return self.values[index]
+        if self.is_list:
```
karlhigley, line 114: We're debating whether it's worth splitting list and non-list columns into two separate classes with a shared base class, but it seemed like more trouble than it's worth at this point. I don't love the `if/else` in all these methods though.

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,149 @@
...
-    A simple wrapper around an array of values
+    A simple wrapper around an array of values. This has an API that's just similar enough to
+    Pandas and cuDF series to be relatively interchangeable from the perspective of the Merlin DAG,
+    but no more. As more methods get added to this class, it gets closer and closer to actually
+    *being* a Pandas/cuDF Series (at which point there's no advantage to using this.) So: keep
+    this class as small as possible.
     """
 
-    def __init__(self, values):
+    def __init__(self, values, row_lengths=None):
```
karlhigley, line 44: We chose to go with row lengths here, but it could equally well be offsets. It's slightly easier to sum the row lengths to get the offsets than to do the subtractions to get the lengths from the offsets, so we did the easy thing here.

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,149 @@
...
 
     @property
     def shape(self):
-        return self.values.shape
+        if self.is_list:
+            dim = self.row_lengths[0] if self.is_ragged else None
+            return (len(self), dim)
+        else:
+            return self.values.shape
+
```
karlhigley, line 141: The `is_list` and `is_ragged` properties below directly align with the schema properties, which should make validating the actual data against the schemas a bit easier than it has been up to this point.

merlin/systems/dag/ops/feast.py
```diff
@@ -298,6 +289,7 @@ def transform(
             Transformed tensor dictionary
         """
         entity_ids = transformable[self.entity_column]
+        array_lib = entity_ids._array_lib
```
karlhigley, line 292: This sets us up to return arrays from the Feast op that match whatever was passed in; if you supply the entity ids as a GPU array, you'll get GPU arrays back.

merlin/systems/triton/conversions.py
```diff
@@ -42,31 +42,98 @@
...
+    Returns
+    -------
+    DictArray
+        Dictionary-like representation of the input columns
+    """
+    dict_inputs = {}
+    for name in column_names:
+        try:
+            values = _get_triton_tensor(request, f"{name}__values")
+            lengths = _get_triton_tensor(request, f"{name}__lengths")
```
karlhigley, line 65: Here we establish a convention for how we expect multi-array features to be formatted in the Triton request. Since Triton can only accept a flat list of named tensors, there's no way for us to group these together without some kind of naming convention.

tests/unit/systems/ops/feast/test_op.py
```diff
@@ -105,29 +106,17 @@ def test_feast_from_feature_view(tmpdir, suffix_int):
                 ColumnSchema(name="prefix_int_feature", dtype=np.int32),
                 ColumnSchema(name="prefix_float_feature", dtype=np.float32),
                 ColumnSchema(
-                    name=f"prefix_int_list_feature_{suffix_int}",
```
karlhigley, line 108: We no longer use suffix integers to identify arrays that go together, since we have the `__values`/`__lengths` convention.

karlhigley (commented):
This doesn't fully solve the problem yet—we still have work to do with the Tensorflow op, for example. But it does move us in the right general direction and make the tests pass.

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,149 @@
...
+from enum import Enum
 from typing import Dict, Optional
 
 import numpy as np
 
-from merlin.core.protocols import SeriesLike, Transformable
+from merlin.core.dispatch import HAS_GPU
+from merlin.core.protocols import SeriesLike
+
+if HAS_GPU:
```
oliverholworthy, line 24: `HAS_GPU` from the dispatch module also requires cudf and dask_cudf are installed. Either replacing this with a try/except ImportError or other method could allow us to run this without these other restrictions.

merlin/systems/dag/dictarray.py
```diff
@@ -13,44 +13,151 @@
...
+        else:
+            return self.values.shape
+
+    @property
+    def is_list(self):
+        return len(self.values.shape) > 1 or self.row_lengths is not None
 
+    @property
+    def is_ragged(self):
+        return self.row_lengths and any(self.row_lengths != self.row_lengths[0])
```
oliverholworthy, line 150: Based on our discussion yesterday. this property is consistent with the data in  `values/row_lengths`. However, for situations where the intermediate data is not ragged according to the row lengths, but does need to be ragged for the purposes of an operator. do we leave that up to an operator to do the coercion?

PR #226, opened 2022-11-14
fil triton runtime operator created and loaded

jperez999:
This PR introduces the triton operator for FIL models. It will allow us to load in specific functionality for FIL when in a triton environment, while allowing us to handle FIL outside of triton environment with different logic.


merlin/systems/dag/runtimes/triton/ops/fil.py
```diff
@@ -0,0 +1,452 @@
...
+            The parameters to pass to the FIL operator.
+        """
+        super().__init__(op)
+        if op is not None:
+            self.fil_op = FILTriton(op.fil_op)
+            self.backend = op.backend
+        if input_schema is not None:
+            self.input_schema = input_schema
+        elif op is not None:
+            self.input_schema = op.input_schema
```
karlhigley, line 67:
```suggestion
        if op is not None:
            self.fil_op = FILTriton(op.fil_op)
            self.backend = op.backend
            self.input_schema = op.input_schema
        if input_schema is not None:
            self.input_schema = input_schema
```

merlin/systems/dag/runtimes/triton/runtime.py
```diff
@@ -30,17 +30,23 @@
 from merlin.core.protocols import Transformable  # noqa
 from merlin.dag import Graph  # noqa
 from merlin.systems.dag.ops import compute_dims  # noqa
+from merlin.systems.dag.ops.fil import PredictForest  # noqa
 from merlin.systems.dag.ops.operator import add_model_param  # noqa
 from merlin.systems.dag.runtimes import Runtime  # noqa
+from merlin.systems.dag.runtimes.triton.ops.fil import PredictForestTriton  # noqa
 from merlin.systems.dag.runtimes.triton.ops.tensorflow import PredictTensorflowTriton  # noqa
 
+# Should this have an import if statement? XGboost, sklearn, CUML???
```
karlhigley, line 39: We might need to figure out a pattern for this, since I don't think we want to make running things in Triton require all the modeling frameworks

karlhigley: Actually, it looks like we have the start of a pattern a few lines farther down 😃

merlin/systems/dag/runtimes/triton/runtime.py
```diff
@@ -30,17 +30,23 @@
...
+
 tensorflow = None
 try:
     import tensorflow
 except ImportError:
     ...
```
karlhigley, line 45: We should probably do a try/except for FIL too

tests/unit/systems/dag/runtimes/test_triton.py
```diff
@@ -77,3 +89,54 @@ def test_triton_runtime_export_and_run(runtime, model_name, expected_model_name,
...
+    ["runtime", "model_name", "expected_model_name"],
+    [
+        (None, None, "ensemble_model"),
+        (TritonEnsembleRuntime(), None, "ensemble_model"),
+        (TritonEnsembleRuntime(), "triton_model", "triton_model"),
+        # (TritonExecutorRuntime(), None, "executor_model"),
+        # (TritonExecutorRuntime(), "triton_model", "triton_model"),
+    ],
+)
+def test_workflow_with_forest_inference(runtime, model_name, expected_model_name, tmpdir):
```
karlhigley, line 105: I'd suggest splitting the specific op tests out into separate files. Maybe a directory like `tests/unit/systems/dag/runtimes/triton/ops`?

merlin/systems/dag/ops/fil.py
```diff
@@ -440,31 +402,39 @@ class XGBoost(FILModel):
...
-        self._learner = learner
-
-        objective = learner["objective"]["name"]
-        if objective == "binary:hinge":
-            raise ValueError(
-                "Objective binary:hinge is not supported."
-                "Only sigmoid and identity values of pred_transform are supported"
-                " for binary classification."
-            )
+        if model:
```
oliverholworthy, line 405: what is the scenario where the model not get passed?

merlin/systems/dag/ops/fil.py
```diff
@@ -318,21 +279,16 @@ def export(
...
-            self.fil_model.num_features,
-            self.fil_model.num_classes,
-            max_batch_size=self.max_batch_size,
-            **self.parameters,
-        )
+        return version_path
 
-        with open(node_export_path / "config.pbtxt", "w", encoding="utf-8") as o:
-            text_format.PrintMessage(config, o)
+    def load_model(self, version_path):
```
oliverholworthy, line 286: what is this load_model required for in the context of the Triton runtimes. Since these models will be inside the FIL  backend, do we need to be able to load them for something else?

merlin/systems/dag/ops/fil.py
```diff
@@ -498,6 +468,14 @@ def save(self, version_path):
         model_path = pathlib.Path(version_path) / self.model_filename
         self.model.save_model(model_path)
 
+    def load(self, version_path) -> "LightGBM":
+        """Save model to version_path."""
+        model_path = pathlib.Path(version_path) / self.model_filename
+        if HAS_GPU:
+            self.model = cuml_fil.load(model_path, output_class=True)
```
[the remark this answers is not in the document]
jperez999: No, according to cuml it can load all cuml, xgboost, lightgbm and treelite, refer here: https://docs.rapids.ai/api/cuml/stable/api.html#cuml.ForestInference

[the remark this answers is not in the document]
jperez999: And treelite can also load lightgbm/xgboost models https://treelite.readthedocs.io/en/latest/treelite-api.html#treelite.Model.load

merlin/systems/dag/ops/fil.py
jperez999: They are not necessary for the triton side, because fil will always run in the triton custom backend. But this is not in the operator for triton, it stayed in the base operator. This will be necessary to load models when we start moving/persisting ensembles around. It is in place for the load_artifacts call that happens in non triton executors. It may be unnecessary for this PR's objective so I can remove, but I wanted to solve for this problem.

merlin/systems/dag/ops/fil.py
jperez999: in when we call from_config, we create a new version of the operator that is not loaded with the model. The model is read in by the config in triton. Which makes it where the only thing the operator does in the triton context is leverage the business logic scripting and sends a request to triton fil model.

karlhigley (approved):
Tests are passing and this looks okay to me, so I'm going to go ahead and approve but refrain from pushing the big green button myself to leave a chance for @oliverholworthy to review it too (if he wants to.)

merlin/systems/dag/ops/compat.py
```diff
@@ -35,6 +39,10 @@
     import cuml.ensemble as cuml_ensemble
 except ImportError:
     cuml_ensemble = None
+try:
+    from cuml import ForestInference as cuml_fil
```
oliverholworthy, line 43: The pattern above with cuml.ensemble and treelite.sklearn was to avoid an error with only importing cuml or treelite. However it might work ok with this.

oliverholworthy (approved):

PR #233, opened 2022-11-22
 Migrate `nvtabular.inference` to `merlin.legacy.inference`

karlhigley:
I don't think we can completely get rid of this code yet, since we're still relying on it to some extent for legacy PyTorch/T4R serving. As we sort out more of the remaining issues that are making it tricky to serve session-based models, we should be able to remove this entirely, but it still needs a home for now. It makes more sense for this code to live in Systems than NVT though, since that consolidates all the Triton code in one repo.

The corresponding PR on the other side is https://github.com/NVIDIA-Merlin/NVTabular/pull/1708


merlin/legacy/inference/triton/data_conversions.py
```diff
@@ -0,0 +1,150 @@
...
+    cudf = cp = None
+
+import numpy as np
+import pandas as pd
+
+from merlin.core.dispatch import build_cudf_list_column, is_list_dtype
+from merlin.dag import Supports
+
+
+def convert_format(tensors, kind, target_kind):
```
oliverholworthy, line 42: Looks like we already have most (all?) of the functions from this file in [merlin.systems.triton.conversions](https://github.com/NVIDIA-Merlin/systems/blob/v0.8.0/merlin/systems/triton/conversions.py#L42)

merlin/legacy/inference/triton/ensemble.py
```diff
@@ -0,0 +1,783 @@
...
+import numpy as np
+import tritonclient.grpc.model_config_pb2 as model_config
+from google.protobuf import text_format
+
+from merlin.core.dispatch import is_string_dtype
+from merlin.schema import Tags
+from nvtabular import ColumnSelector
+
+
+def export_tensorflow_ensemble(
```
oliverholworthy, line 30: similar to the data conversions module, looks like we already have these in [merlin.systems.triton.export](https://github.com/NVIDIA-Merlin/systems/blob/v0.8.0/merlin/systems/triton/export.py#L34)

oliverholworthy:
We appear to already have most of this in systems already apart from the benchmarking tools. There are some differences between the files (the change of nnzs to lengths for example).


| Legacy                                                | Systems                                        | Difference                                                           |
|-------------------------------------------------------|------------------------------------------------|----------------------------------------------------------------------|
| `merlin/legacy/triton/__init__.py`                      | `merlin/systems/triton/__init__.py`              | systems has changed signature of convert_df_to_triton_input function |
| `merlin/legacy/triton/benchmarking_tools.py`            | -                                                | MISSING                                                              |
| `merlin/legacy/triton/data_conversions.py`              | `merlin/systems/triton/conversions.py`           | systems has additional functions for dict array conversions          |
| `merlin/legacy/triton/ensemble.py`                      | `merlin/systems/triton/export.py`                | nnzs -> lengths, the way dims are computed (sparse_max)              |
| `merlin/legacy/inference/triton/model/model_pt.py`      | `merlin/systems/triton/models/pytorch_model.py`  | nnzs -> lengths                                                      |
| `merlin/legacy/inference/triton/workflow/base.py`       | `merlin/systems/workflow/base.py`                | none                                                                 |
| `merlin/legacy/inference/triton/workflow/hugectr.py`    | `merlin/systems/workflow/hugectr.py`             | none                                                                 |
| `merlin/legacy/inference/triton/workflow/pytorch.py`    | `merlin/systems/workflow/pytorch.py`             | some changes to tensor transformations relating to sparse tensors    |
| `merlin/legacy/inference/triton/workflow/tensorflow.py` | `merlin/systems/workflow/tensorflow.py`          | nnzs -> lengths                                                      |
| `merlin/legacy/inference/triton/workflow_model.py`      | `merlin/systems/triton/models/workflow_model.py` | nnzs -> lengths                                                      |

There are also some duplicate workflow files between `merlin/legacy/inference/triton/workflow` `merlin/legacy/inference/workflow`,

merlin/legacy/inference/triton/data_conversions.py
karlhigley: We do, but this is intended to be a direct copy/paste into a standalone package that's easy to delete. Since this is for maintaining legacy support, I don't want to consolidate the repetitive functions, because changing the versions we're actively using in Systems should be okay but changing these may break legacy functionality that has limited tests.

merlin/legacy/inference/triton/ensemble.py
karlhigley: Yeah, a lot of this is there, but trying to figure out what is and isn't the same, which parts of the Systems code have changed/evolved, etc has been stopping us from removing the old code from NVT. Putting the old code from NVT verbatim in a completely separate standalone package bypasses that concern. If we care enough to do the work of figuring it out and consolidating we can do that, but my plan is to let this code sit here for a few months while we sort out session-based serving and then delete the whole package.

karlhigley:
The changes you've listed aren't necessarily backwards compatible, which is why adding a whole additional copy of this stuff makes a certain amount of sense to me. It can be dropped in a separate package, marked as legacy, and left alone until we can completely remove it from Merlin in (hopefully) a few months. We could even set code owners on the folder (e.g. me) so that changes to it can't be made accidentally without someone keeping an eye out to make sure that doesn't happen.

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

karlhigley:
Still want to get the old inference code out of NVTabular, but this may not be the right destination, given how confusing it's been to have two different versions of the serving code in this library that run on different back-ends.

PR #237, opened 2022-11-29
update triton to allocate more cuda memory to pool size

jperez999:
This will help stabilize the triton runtime fil op tests. It allows triton to allocate ~500MB instead of the default 64MB. This helps when loading models specifically FIL backend models.


jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

PR #238, opened 2022-11-29
Nvt op runtime triton

jperez999:
Introduces the triton runtime version of the nvtabular operator. Allows for hot-swapping of the nvtabular operator from ensemble depending on the target runtime.


merlin/systems/dag/ops/workflow.py
```diff
@@ -69,68 +61,19 @@ def __init__(
...
         self.model_framework = model_framework or ""
         self.cats = cats or []
         self.conts = conts or []
-        self._python = backend == "python"
 
-        if workflow is not None:
-            self.input_schema = workflow.input_schema
-            self.output_schema = workflow.output_schema
-
-    def transform(self, col_selector: ColumnSelector, transformable: Transformable):
```
jperez999, line 85: removed some of the basic methods because they require refactor for the local runtime. And this PR focuses on the triton runtime migration. Will readd this functionality when we start working on local runtimes. Meanwhile the basic classes will rely on base class function definitions which raise not implemented errors.

tests/unit/systems/ops/nvtabular/test_op.py
```diff
@@ -29,7 +29,8 @@
 from nvtabular import ops as wf_ops  # noqa
 
 ensemble = pytest.importorskip("merlin.systems.dag.ensemble")
-workflow_op = pytest.importorskip("merlin.systems.dag.ops.workflow")
+wf_op = pytest.importorskip("merlin.systems.dag.ops.workflow")
```
jperez999, line 32: Should this be moved to the triton ops directory, since it tests the triton op behaviors?

karlhigley: Yeah, that's probably a good idea 👍🏻

merlin/systems/dag/ops/workflow.py
```diff
@@ -13,19 +13,11 @@
...
 
-from merlin.core.protocols import Transformable  # noqa
 from merlin.dag import ColumnSelector
-from merlin.schema import ColumnSchema, Schema
-from merlin.systems.dag.ops.operator import PipelineableInferenceOperator  # noqa
-from merlin.systems.triton.conversions import (
-    dict_array_to_triton_request,
-    triton_response_to_dict_array,
-)
-from merlin.systems.triton.export import generate_nvtabular_model
```
karlhigley, line 28: Love to see this clean-up happening. You can tell how much decoupling is happening just by reading the imports. ❤️

merlin/systems/dag/ops/workflow.py
```diff
@@ -156,47 +99,4 @@ def export(
...
-        -------
-        String
-            Name of the current class as spelled in module.
-        """
-        cls_name = self.__class__.__name__.lower()
-        if self._python:
-            return f"py_{cls_name}"
-        else:
-            return cls_name
+        raise NotImplementedError
```
karlhigley, line 102: I assume this is required because the base class would go ahead an export if we didn't raise an error here, right? 👍🏻

karlhigley (approved):
Aside from moving that test, LGTM

PR #242, opened 2022-12-01
Update API and add notebook

mikemckiernan:
- Update the API docs to include newer Operators, exporting functions for ensembles, and conversion functions for Triton Inference Server.

- Add the notebook that demonstrates how to serve an implicit model.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -25,9 +25,23 @@ Ensemble Operator Constructors
 
    workflow.TransformWorkflow
    tensorflow.PredictTensorflow
+   fil.PredictForest
+   implicit.PredictImplicit
    softmax_sampling.SoftmaxSampling
    session_filter.FilterCandidates
    unroll_features.UnrollFeatures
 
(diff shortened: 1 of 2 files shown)


docs/source/api.rst
```diff
@@ -25,9 +25,36 @@ Ensemble Operator Constructors
...
+-------------------------------
+
+.. currentmodule:: merlin.systems.triton
+
+.. autosummary::
+   :toctree: generated
+
+   export_tensorflow_ensemble
+   export_pytorch_ensemble
+   export_hugectr_ensemble
```
karlhigley, line 48: I don't think we actually want to include these three functions in the docs, since they're legacy code inherited from NVT that we'd like to get rid of 😅

mikemckiernan: NP, I'll ditch them.

mikemckiernan:
rerun tests

bschifferer (approved):

bschifferer (approved):

PR #246, opened 2022-12-04
working local pytorch transform

jperez999:
This PR adds the transform logic required to run an inference call on a trained pytorch model without triton in an ensemble graph.

### DIFF
--- a/merlin/systems/dag/ops/pytorch.py
+++ b/merlin/systems/dag/ops/pytorch.py
@@ -15,11 +15,14 @@
 #
 import os
 
+import numpy as np
 import torch  # noqa
 import torch.utils.dlpack  # noqa
 
+from merlin.core.protocols import Transformable  # noqa
 from merlin.dag import ColumnSelector  # noqa
 from merlin.schema import Schema  # noqa
+from merlin.systems.dag.dictarray import DictArray
 from merlin.systems.dag.ops.operator import PipelineableInferenceOperator  # noqa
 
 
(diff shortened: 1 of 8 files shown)


merlin/systems/dag/ops/pytorch.py
```diff
@@ -93,6 +96,17 @@ def compute_output_schema(
         """
         return self.output_schema
 
+    def transform(self, col_selector: ColumnSelector, transformable: Transformable):
+        tensor_dict = {}
+        for column in transformable.columns:
+            tensor_dict[column] = torch.from_numpy(np.squeeze(transformable[column].values))
```
jperez999, line 102: Should we be doing this in another place? Should dict array have APIs for transforming to tensors of specific frameworks?

merlin/systems/dag/ops/pytorch.py
```diff
@@ -93,6 +96,17 @@ def compute_output_schema(
...
+
+        result = self.model(tensor_dict)
+        output = {}
+        for idx, col in enumerate(self.output_schema.column_names):
```
jperez999, line 106: This allows for more than one output, but ordering will matter at this point. Is there a better way to go about handling a multi output where we can disregard order? Luckily dictionaries hold on to order...

merlin/systems/dag/ops/pytorch.py
karlhigley: I think it should eventually (soon, even) but since we don't have those yet, this seems reasonable for now

merlin/systems/dag/ops/pytorch.py
karlhigley: In this case, it seems like letting the model return a dictionary is okay, since we don't have the limitations that the Triton Torch backend imposes on us 🤔

tests/unit/systems/dag/runtimes/local/ops/torch/test_op.py
```diff
@@ -0,0 +1,140 @@
...
+  dims: 1
+}
+parameters {
+  key: "INFERENCE_MODE"
+  value {
+    string_value: "true"
+  }
+}
+backend: "pytorch"
+"""
```
karlhigley, line 81: Doesn't look like this is used:
```suggestion
```

karlhigley (approved):

tests/unit/systems/dag/runtimes/local/ops/torch/test_op.py
jperez999: removed

jperez999:
rerun tests

karlhigley:
rerun tests

PR #247, opened 2022-12-06
Dictarray to and from df

jperez999:
This PR isolates the integration of functionality into dict array that allows it to support transforming to and from a dataframe.

### DIFF
--- a/merlin/systems/dag/dictarray.py
+++ b/merlin/systems/dag/dictarray.py
@@ -230,9 +230,9 @@ class DictArray:
         """
         Create a DataFrame from the DictArray
         """
-        df = get_lib.DataFrame()
+        df = get_lib().DataFrame()
         for col in self.columns:
-            df[col] = get_lib.Series(self[col])
+            df[col] = get_lib().Series(self[col])
         return df
 
     @classmethod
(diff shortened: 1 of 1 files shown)


jperez999:
rerun tests

nv-alaiacano (approved):

merlin/systems/dag/dictarray.py
```diff
@@ -225,6 +226,25 @@ def copy(self):
...
+        return df
+
+    @classmethod
+    def from_df(cls, df):
+        """
+        Create a DictArray from a DataFrame
+        """
+        array_dict = {}
+        for col in df.columns:
+            array_dict[col] = df[col].to_numpy()
```
oliverholworthy, line 245: We might need some extra logic here to handle list values. There's currently a bug in cudf that results in this throwing a `TypeError` if we hit this line with a series with a list dtype.

https://github.com/rapidsai/cudf/issues/12115

```python
import cudf
cudf.Series([[1]]).to_numpy()
# => Raises TypeError: issubclass() arg 1 must be a class
```

And in the case of cudf would we want to keep the data on the same device when using this method?

jperez999: This is a mini step that allows us to get the current tests working. We will not be keeping this in the long term. It will be modified once merlin dtypes and columns are in place in Core.

PR #249, opened 2022-12-06
Nvt local runtime

jperez999:
This allows for the use of nvtabular workflows with the local runtime. it requires changes to core in the PR: https://github.com/NVIDIA-Merlin/core/pull/180


karlhigley (approved):

karlhigley:
It seems like this is hitting dtype issues related to the difference between actual scalars and lists of one element. Wondering what it would take to make a clearer distinction between those two cases. Do we have to sort out shapes?

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

PR #252, opened 2022-12-12
Restore Support for NVTabular Workflow + PyTorch Model Export

oliverholworthy:
## Goal

Restores support for NVTabular Workflow + PyTorch ensemble with `merlin.systems.triton.export.export_pytorch_ensemble`

## Implementation Details

- ragged inputs/outputs not currently handled correctly. Updating the Triton config to match the tensor shapes as 1D arrays.
- Using the `output_model` parameter of `merlin.systems.triton.export.generate_nvtabular_model` to configure which Workflow class is used to apply post-processing of the worflow transformed data.
- Corrects the path to the pytorch model file so that the `export_pytorch_ensemble` runs and picks up the correct python model.
- `pytorch_model.py` - handle a tensor output from the model (now the default output of Transformers4Rec - this changed recently) as well as a dictionary of tensors.
- the helper function to create input data for requests was passing offsets instead of row lengths into the `__lengths` input tensor


karlhigley:
@oliverholworthy Is this actually being used somewhere? AFAIK, we're using the version of `export_pytorch_ensemble` in NVTabular. The version of that code living here got copied over when Systems was created and exists here as somewhat of an artifact of the development process. (I don't object to patching it up, just want to make sure we're on the same page.)

oliverholworthy:
@karlhigley I arrived at this PR from the goal of trying to get an operator-based ensemble of NVTabular Workflow + Transformers4Rec PyTorch model working together. And was using this legacy function as a comparison. And while looking at that, noticed that some changes were required to make that work.

Once we get the operator version working we won't need this function. However for now, maybe it's worth getting it running again. Then we maybe don't need to merge #233

karlhigley:
@oliverholworthy The version of the code in NVT is mostly frozen in a working configuration; the version in this library wasn't ever intended to be used, it just served as a starting point for refactoring in the direction of the operator API. Since the legacy inference code didn't have great tests, there's not a great way to make sure that code keeps working without either freezing it in place (like in NVT) or wrapping it with new tests (here in Systems.) If you want to pursue the latter to build a working version of the old code in this library, I won't stand in your way, but it seemed like a lot of work to fix code we don't recommend that anyone uses. 🤷🏻 

It sounds like you might be aiming to remove the legacy inference code in NVT though? That would be cool!

karlhigley:
@rnyak @angmc @nv-alaiacano With Oliver out, I'm trying to work backwards to figure out what the original need was and if there's another way to fulfill it. I suspect one of y'all might know: Was this for examples, GTC, AI Workflows, or all/none of the above?

PR #255, opened 2022-12-15
changing the default runtime from Ensemble to Executor

jperez999:
This PR swaps the default runtime from Ensemble to Executor. And removed Ensemble Runtime from systems completely.


karlhigley:
In order to address some of the cross-repo shape-related issues that are cropping up, we need to be able to run some additional code around model predictions, which there's no good way to do in ensemble mode. So, at this point, since we're ready to move away from ensemble mode and toward executor mode anyway, we can remove ensemble mode entirely. It won't get us where we want to go, and it's always available in the commit history if we want to go back and look at or revive it someday.

karlhigley (approved):

karlhigley:
Looks like there's a place where the tests assume the name of the exported model—should be a quick fix

PR #265, opened 2022-12-29
Remove unused legacy serving code

karlhigley:
This code was copied from NVTabular with the intention of using it as a starting point for refactoring in the direction of an operator-based serving API (which now mostly exists), but the remaining unused code was never removed. This PR removes it an effort to avoid confusing users with multiple versions of the legacy serving code they could (attempt to) use.

See also:
* A previous attempt to remove the legacy serving code from NVT and move it to Systems wholesale:
  * https://github.com/NVIDIA-Merlin/NVTabular/pull/1708
  * https://github.com/NVIDIA-Merlin/systems/pull/233
* An effort to fix the legacy code in this repo (which isn't generally expected to be used):
  * https://github.com/NVIDIA-Merlin/systems/pull/252
* An alternative approach to maintaining the legacy serving functionality without moving it to Systems:
  * https://github.com/NVIDIA-Merlin/serving/pull/2


karlhigley:
This shouldn't cause downstream issues, because this library doesn't seem to have users yet, because these methods have been omitted from our documentation, because we haven't (AFAIK) been using them in examples or otherwise doing anything to indicate that they exist or should be used. Whether this constitutes a breaking change or not is a semantic issue I don't care to debate at any length, but please understand that avoiding changes to the legacy serving code that would impact our users is why this code existed here (in addition to in NVT) in the first place.

merlin/systems/dag/runtimes/triton/ops/workflow.py
```diff
@@ -13,18 +13,27 @@
...
 # limitations under the License.
 #
+import importlib.resources
 import json
+import os
 import pathlib
+from shutil import copyfile
+
+# this needs to be before any modules that import protobuf
+os.environ["PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION"] = "python"
```
jperez999, line 23: we can remove this and then we can remove all the # noqa comments.

jperez999 (approved):

PR #278, opened 2023-01-24
refactor systems to handle shape in columns

jperez999:
This PR adds in support for shape in column schemas. It removes the use of optional arguments is_list and is_ragged. Now those arguments are used in the properties dictionary to be consumed by the column schema.

Depends on https://github.com/NVIDIA-Merlin/core/pull/195


karlhigley:
Superseded by subsequent work on schema shapes and `TensorTable`

PR #279, opened 2023-01-25
refactor out pipelineable inference operator

jperez999:
This PR removes the pipelineable inference operator and moves that functionality into the TritonOperator. This minimizes the number of sub classes available to the inference operator and better organizes the sub classes into their respective behaviors (i.e. TritonOperator for triton and inference operator for non triton logic).


tests/unit/systems/utils/ops.py
```diff
@@ -20,10 +20,13 @@
...
 from merlin.systems.dag import DictArray
 
-inf_op = pytest.importorskip("merlin.systems.dag.ops.operator")
+inf_op = pytest.importorskip("merlin.systems.dag.runtimes.triton.ops.operator")
 
 
-class PlusTwoOp(inf_op.PipelineableInferenceOperator):
+class PlusTwoOp(inf_op.TritonOperator):
+    def __init__(self):
+        super().__init__(self)
```
karlhigley, line 28: This looks like it might be unnecessary, but I could be missing something 🤔

karlhigley (approved):

merlin/systems/dag/runtimes/triton/ops/feast.py
```diff
@@ -0,0 +1,345 @@
...
+    The QueryFeast operator is responsible for ensuring that your feast feature store [1]
+    can communicate correctly with tritonserver for the ensemble feast feature look ups.
+
+    References
+    ----------
+    [1] https://docs.feast.dev/
+    """
+
+    @classmethod
+    def from_feature_view(
```
karlhigley, line 34: It seems like a lot of this is the same as the definition of the method in the base class, so you might be able to call `super()` and then wrap the result with `QueryFeastTriton`

jperez999: The super class here is the TritonOperator, I am not sure that has from_feature_view? I would love to have the base operator as a parent class. Can I put more than one operator to inherit from? What would supersede what?

karlhigley (approved):

PR #280, opened 2023-01-26
Add error messages for NotImplementedError Exceptions

jperez999:
This PR provides better NotImplementedErorr messaging. It gives the user more information around accessing certain functionality and what types of operators might have that functionality.


merlin/systems/dag/ops/operator.py
```diff
@@ -114,7 +117,11 @@ def export(
         Node_configs: list
             A list of individual configs for each step (operator) in graph.
         """
-        raise NotImplementedError
+        raise NotImplementedError(
+            "Export function only exists for the specific runtimes, that use this function"
+            f" (i.e. Triton). Please replace {self.__class__.__name__} with the appropriate "
+            f"runtime specific operator, possibly {self.__class__.__name__}Triton"
+        )
```
karlhigley, line 124: This looks pretty good! Wordsmithing a little bit, how about something like:
```
Exporting an operator to run in a particular context (e.g. Triton) only makes sense when a runtime is specified. To select an operator for the appropriate runtime, replace {self.__class__.name} with a runtime-specific operator  class, possibly {self.__class__.name}Triton.
```

merlin/systems/dag/ops/operator.py
```diff
@@ -62,7 +62,10 @@ def transform(
         DataFrame
             Returns a transformed dataframe for this operator
         """
-        raise NotImplementedError
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not have a Transform function."
+            "Please create one, if required."
```
karlhigley, line 67: I think I might be confused about what this is telling me to do if I saw this error message. It seems like one of the important details to include might be about base operators vs runtime operators? Like...where should I add a `transform` method if I see this error?

merlin/systems/dag/ops/operator.py
jperez999: I am wondering if we should even require the export function in the Base Operator. Currently only runtime specific operators would possible require an export function, thoughts?

karlhigley: Yeah, I wonder about removing it too. I think I'm good with that, but I wonder if we should then move `export()` to a `RuntimeOperator` base class or something?

jperez999: So essentially we would be relocating the pipelineable inference operator to a RuntimeOperator? Instead of wanting to refactor it out, like we did in https://github.com/NVIDIA-Merlin/systems/pull/279.

karlhigley (approved):

PR #288, opened 2023-02-13
Update `compute_dims` function to use new shape property

oliverholworthy:
Update `compute_dims` function to use new shape property.

### DIFF
--- a/merlin/systems/dag/ops/__init__.py
+++ b/merlin/systems/dag/ops/__init__.py
@@ -38,10 +38,11 @@ def compute_dims(col_schema, scalar_shape=None):
     assert isinstance(column_dims, list)
 
     if col_schema.is_list:
-        value_count = col_schema.properties.get("value_count", None)
-        if value_count and value_count["max"] > 0 and value_count["min"] == value_count["max"]:
-            column_dims = [value_count["max"]]
-        else:
-            column_dims = [-1]
+        column_dims = []
+        for dim in col_schema.shape.dims[1:]:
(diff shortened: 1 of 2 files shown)


karlhigley (approved):

merlin/systems/dag/ops/softmax_sampling.py
```diff
@@ -99,7 +99,7 @@ def compute_output_schema(
         self, input_schema: Schema, col_selector: ColumnSelector, prev_output_schema: Schema = None
     ) -> Schema:
         """Describe the operator's outputs"""
-        return Schema([ColumnSchema("ordered_ids", dtype=np.int32, is_list=True, is_ragged=False)])
+        return Schema([ColumnSchema("ordered_ids", dtype=np.int32, dims=(None, 1))])
```
oliverholworthy, line 102: We now need to specify the second dimensiion output size as 1 so that the output is fixed size and returns a single `ordered_ids` in the Triton config instead of split into values/lengths

oliverholworthy: `(None, 1)` matches the return shape of `(-1, 1)` from the transform method. There may be some follow-up work to make this operator work on more than one batch at a time instead of only the one that it assumes currently.

karlhigley: Good catch! Agreed on the follow-up work, that’s been on our list for a while but we decided it would be easier to tackle after we did the shapes migration.

PR #297, opened 2023-03-17
update conftest for backwards compat and new to parquet api

jperez999:
For cudf version 22.10.00+ the to_parquet api has changed. It no longer support the chunk_size call. It was replaced with row_group_size_rows, which can take minimum of 5000 rows. We have updated conftest call to reflect this change while maintaining backwards compatibility. We need this for all future versions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -152,9 +152,25 @@ def datasets(tmpdir_factory):
 
     half = int(len(df) // 2)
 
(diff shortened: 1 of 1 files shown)


karlhigley:
This has the same CUDA initialization error failures in the tests that we've been seeing on other PRs, so I'm going to merge it anyway

PR #298, opened 2023-03-17
Fix Tensorflow serving to make Merlin example tests pass again

karlhigley:


tests/unit/systems/utils/tf.py
```diff
@@ -22,33 +22,44 @@
...
+        for input_name, input_ in inputs.items():
+            reshaped_inputs[input_name] = tf.reshape(input_, (-1, 1))
+
+        x_emb_output = self.emb_layer(reshaped_inputs)
+        x = self.dense_layer(x_emb_output)
+        x = self.output_layer(x)
+        return x
+
+
+def create_tf_model(cat_columns: list, cat_mh_columns: list, embed_tbl_shapes: dict):
```
oliverholworthy, line 58: how much of this is a refactor vs fucntional change? It appears to be mostly a change to the input shape that's going on here. Which may be possible to make on the previous version with a change to the `shape` argument of the Keras Input

karlhigley: I couldn't find any way to get this to work with the functional API and it reportedly won't work with the Sequential API either

merlin/systems/triton/conversions.py
```diff
@@ -121,10 +121,15 @@ def dict_array_to_triton_request(model_name, dictarray, input_col_names, output_
     """
     input_tensors = []
 
-    for name, column in dictarray.items():
+    for name, column in tensor_table.items():
         if name in input_col_names:
-            col_tensor = _triton_tensor_from_array(name, column.values)
-            input_tensors.append(col_tensor)
+            if column.is_ragged:
+                values = _triton_tensor_from_array(f"{name}__values", column.values)
```
oliverholworthy, line 127: I'm wondering if it would be helpful to have some TensorTable -> Triton Tensor conversion built into core somehow (not blocking this PR). Or perhaps some way to  extend core with registered conversons here or part of another package?

oliverholworthy (approved):
Only thing missing I think is correcting the output name in the TensorFlow example model from the default `output_1` (returning a dict from the call method may with that)

merlin/systems/triton/conversions.py
karlhigley: Yeah, that all sounds good to me. Not sure exactly what it should look like yet either though

karlhigley:
@oliverholworthy I tried to change the places that use `output_1` that are based on TF models we create explicitly in the tests, but we probably will still need to make upstream updates to Merlin Models to make those models match

PR #305, opened 2023-03-27
Remove `ensemble` specific code 

oliverholworthy:
Following the removal of the ensemble runtime export in #255. Further cleanup of methods that are no longer required.  

## Motivation

- Make it clearer which operators are required to implement for those creating new or custom operators. Cleaning up unused methods from our operators, will hopefully make them easier to use as a reference and get started creating operators.

## Details

One new feature is added by this PR. A `save_artifacts` method that complements the `load_artifacts` method. And facilitates the removal of the `export` methods on all non-Triton operators.

- Removes `OperatorRunner` (and corresponding Triton Python Model `oprunner_model.py`)
  - This has been replaced by the `TritonExecutorRuntime` (and corresponding Triton Python Model `executor_model.py`)
  - Removes `from_config` methods from non-Triton specific operators (this method was called by `oprunner_model.py` to reload an operator from the Triton model config). This was used when we were exporting each operator as a triton model)
- Removes unused `export` methods from non-Triton specific operators.
  - These methods were used when we had the 'ensemble' runtime where each operator exported to their own Triton model
 - Removes `exportable_backends` from operators
   - We now only have one type of Triton Runtime. Since all the Triton operators require an export method so we know it's exportable if the operator is a `TritonOperator`.
   - Some non-Triton operators were implementing the `export` to save non-pickleable objects for reloading. Added a new method called `save_artifacts` that serves this purpose and is intended to be implemented on any operator that implements `load_artifacts`.
 - Removes `PredictImplicitTriton`. The regular `PredictImplicit` is sufficient now that we have `save_artifacts` implemented.


merlin/systems/dag/runtimes/triton/ops/operator.py
```diff
@@ -107,6 +67,7 @@ def transform(
         """
         return transformable
 
+    @abstractmethod
```
oliverholworthy, line 70: This `export` method of the `TritonOperator` is now an abstractmethod. Meaning that an operator that inherits from TritonOperator is required to implement the export method.

merlin/systems/dag/runtimes/triton/runtime.py
```diff
@@ -220,11 +213,11 @@ def _executor_model_export(
         return config
 
 
-def _create_node_table(nodes, backend):
+def _create_node_table(nodes):
     exportable_node_idx = 0
     node_id_lookup = {}
     for node in nodes:
-        if node.exportable(backend):
+        if isinstance(node.op, TritonOperator):
```
oliverholworthy, line 220: We don't need to check the `exportable` method anymore because every `TrtionOperator` is expected to implement `export` to save the operator as a Triton model.

karlhigley (approved):

PR #310, opened 2023-04-01
Fix `PredictTensorflow` args in session-based serving integration test

karlhigley:


jperez999 (approved):
This is much cleaner!

PR #316, opened 2023-04-06
Update Triton Workflow runner to handle ragged outputs

oliverholworthy:
Part of https://github.com/NVIDIA-Merlin/Merlin/issues/255
Follow-up to https://github.com/NVIDIA-Merlin/systems/pull/314

Updates `_transform_outputs` in `TensorflowWorkflowRunner` to handle the case where a workflow output is ragged. And we are expecting to output `{col_name}__values` / `{col_name}__offsets` even when the row lengths have the same length (for example, when we only have one row)

### DIFF
--- a/merlin/systems/triton/conversions.py
+++ b/merlin/systems/triton/conversions.py
@@ -298,11 +298,15 @@ def _pandas_to_array(df, cpu=True):
     for name in df.columns:
         col = df[name]
         if pd.api.types.is_list_like(col.values[0]):
-            offsets = pd.Series([0]).append(col.map(len).cumsum()).values
-            if not cpu:
-                offsets = cp.array(offsets)
             values = array_type(list(itertools.chain(*col)))
-            output[name] = (values, offsets)
+            row_lengths = col.map(len)
+            if all(row_lengths == row_lengths[0]):
+                output[name] = values.reshape((-1, row_lengths[0]))
+            else:
+                offsets = pd.Series([0]).append(row_lengths.cumsum()).values
+                if not cpu:
+                    offsets = cp.array(offsets)
+                output[name] = (values, offsets)
         else:
             values = col.values
             if not cpu:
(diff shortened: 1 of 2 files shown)


merlin/systems/workflow/tensorflow.py
```diff
@@ -56,13 +56,27 @@ def _transform_outputs(self, tensors):
                 d = d.reshape(row_dim, col_dim)
                 output_tensors.append((name, d))
             elif isinstance(value, tuple):
-                # convert list values to match TF dataloader
-                values = value[0].astype(self.output_dtypes[name])
```
oliverholworthy, line 60: In the case of a Workflow with a ragged output. this output_dtypes contains the keys matching the triton output names e.g. `{col_name}__values`. This was raising a KeyError here in this lookup into output_dtypes.

Added a condition to check the column schema `is_ragged` attribute to conditionally return the ragged outputs

merlin/systems/triton/conversions.py
```diff
@@ -321,14 +317,9 @@ def _cudf_to_array(df, cpu=True):
...
-            values = col.list.leaves.values_host if cpu else col.list.leaves.values
             offsets = col._column.offsets.values_host if cpu else col._column.offsets.values
-
-            row_lengths = offsets[1:] - offsets[:-1]
-            if all(row_lengths == row_lengths[0]):
-                output[name] = values.reshape((-1, row_lengths[0]))
-            else:
-                output[name] = (values, offsets)
+            values = col.list.leaves.values_host if cpu else col.list.leaves.values
+            output[name] = (values, offsets)
```
karlhigley, line 322: This actively rolls back some of the changes that made the T4R notebook work. I don't think we should use values/offsets for all list features, just the ragged ones.

merlin/systems/workflow/tensorflow.py
```diff
@@ -56,13 +56,27 @@ def _transform_outputs(self, tensors):
...
                 offsets = value[1].astype(np.int32)
-                output_tensors.append((name + "__offsets", offsets))
+                if self.workflow.output_schema[name].is_ragged:
+                    values = values.astype(self.output_dtypes[name + "__values"])
+                    output_tensors.append((name + "__values", values))
+                    output_tensors.append((name + "__offsets", offsets))
+                else:
+                    row_lengths = offsets[1:] - offsets[:-1]
+                    if not all(row_lengths == row_lengths[0]):
+                        raise ValueError(
```
karlhigley, line 68: I like this validation 👍🏻

merlin/systems/triton/conversions.py
oliverholworthy: It seemed easier to do it this way, rather than trying to conditionally convert from a fixed size tensor back into the ragged represntation in the case where we have one row or happen to have rows of the same length (but still want to output the ragged representation to match the workflow output schema and the corresponding triton model

PR #317, opened 2023-04-06
Convert Workflow outputs to Triton-friendly format using `TensorTable`

bschifferer:
Fix some bug related to serving NVT + TensorFlow with Ragged Tensors.

This is an alternative to https://github.com/NVIDIA-Merlin/systems/pull/316


karlhigley (approved):

karlhigley:
This is a breaking change because it removes support for converting ragged Workflow outputs to dense via `sparse_max`. I think we should go ahead with it, because we're currently in the process of replacing `sparse_max` with an actual padding operator anyway.

PR #319, opened 2023-04-07
Condense the `WorkflowRunner` sub-classes

karlhigley:
The workflow output format should now be the same for both TF and Torch, so we can condense their respective sub-classes into the base `WorkflowRunner`, which only leaves the HugeCTR sub-class to deal with later (when we have HugeCTR support.)

This is a follow-up to @bschifferer's [PR that cleaned up the `TensorflowWorkflowRunner` so much it became trivial](https://github.com/NVIDIA-Merlin/systems/pull/317).


karlhigley:
Unclear to me why the GPU test run was marked as failing, because the tests all passed 🤷🏻

jperez999 (approved):

PR #320, opened 2023-04-07
Use `Runtimes` and `Executors` to serve NVT `Workflow`s

karlhigley:
This reworks the existing NVTabular Workflow serving to use some of the newer concepts we've introduced, while continuing to support the optimized C++ implementations of NVT ops for serving:
* It creates a sub-class of `LocalExecutor` that contains the custom code for running Workflows at serving time (which involves some additional conversion between data formats)
* It creates a sub-class of `Runtime` that contains the appropriate operator implementations substitutions
* It creates a small `OpTable` abstraction in order to encapsulate logic about when ops should be substituted

Depends on NVIDIA-Merlin/core#279


merlin/systems/dag/runtimes/base_runtime.py
```diff
@@ -33,9 +34,32 @@ def __init__(self, executor=None):
...
+        ----------
+        graph : Graph
+            Graph of nodes container operator chains for data manipulation.
+
+        Returns
+        -------
+        Graph
+            Copy of the graph with operators converted to this runtime's versions
+        """
+        if self.op_table:
```
oliverholworthy, line 53: if `self.op_table` is always an instance of `OpTable` this contition might always be true

merlin/systems/dag/runtimes/base_runtime.py
```diff
@@ -44,12 +68,17 @@ def transform(self, graph: Graph, transformable: Transformable):
             Graph of nodes container operator chains for data manipulation.
         transformable : Transformable
             Input data to transform in graph.
+        convert: bool
```
oliverholworthy, line 71: what's the scenario where we'd want to pass `convert=False`?

merlin/systems/dag/runtimes/base_runtime.py
karlhigley: Ah, good catch. A plain dictionary would evaluate as false-y if it was empty here. 👍🏻

merlin/systems/dag/runtimes/nvtabular/executor.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # Convert the whole list into the same format
+        formatted_tensors = []
+        for upstream_tensors in node_input_data:
+            upstream_tensors = _convert_format(upstream_tensors, target_format)
+            formatted_tensors.append(upstream_tensors)
+
+        return formatted_tensors
+
+
```
karlhigley, line 138: Note that pretty much everything in this file from here down is just getting moved from other places. There's definitely more refactoring we want to do on the code below, but moving it here more clearly communicates that the usage of this code is isolated to this single file and can be freely refactored without causing issues elsewhere.

merlin/systems/dag/runtimes/nvtabular/executor.py
```diff
@@ -0,0 +1,305 @@
...
+    elif data.column_type is NumpyColumn:
+        return Supports.CPU_DICT_ARRAY
+    else:
+        if isinstance(data, TensorTable):
+            raise TypeError(f"Unknown type: {data.column_type}")
+        else:
+            raise TypeError(f"Unknown type: {type(data)}")
+
+
+def _convert_format(tensors, target_format):
```
karlhigley, line 185: A next step along the road of this refactor is to replace `DictArray` with `TensorTable` and get rid of the manual conversions below. I think we'll still need a function like `_convert_format`, but the actual conversions it uses should be the dataframe/`TensorTable` conversions from Merlin Core.

merlin/systems/dag/runtimes/nvtabular/executor.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def _execute_node(self, workflow_node, input_tensors, capture_dtypes=False, strict=False):
+        upstream_outputs = self._run_upstream_transforms(workflow_node, input_tensors)
+        upstream_outputs = self._merge_addl_root_columns(
+            workflow_node, input_tensors, upstream_outputs
+        )
+        tensors = self._standardize_formats(workflow_node, upstream_outputs)
+
+        transform_input = _concat_tensors(tensors)
+        # TODO: In order to replace the line above with the line below, we first have to replace
```
oliverholworthy, line 94: This is one difference from the LocalExecutor._execute_node. The other seems to be the second line calling `_merge_addl_root_columns` instead of `_append_addl_root_columns`. Is that something that we might be able to unify with `LocalExecutor` eventually too?

merlin/systems/dag/runtimes/nvtabular/executor.py
```diff
@@ -0,0 +1,305 @@
...
+from merlin.core.compat import pandas
+from merlin.core.dispatch import build_cudf_list_column, concat_columns, is_list_dtype
+from merlin.dag import Graph, Node, Supports
+from merlin.dag.executors import LocalExecutor
+from merlin.table import CupyColumn, NumpyColumn, TensorTable
+
+LOG = logging.getLogger("merlin-systems")
+
+
+class NVTabularServingExecutor(LocalExecutor):
```
oliverholworthy, line 32: Is the idea that we're working toward unifying the functionality required here into `LocalExecutor` entirely?

merlin/systems/dag/runtimes/nvtabular/executor.py
karlhigley: Yeah, we're aiming to [address those things](https://github.com/NVIDIA-Merlin/systems/pull/328/commits/e282682ac2e16b66444cee2f5d8f0e9acbbc1697) in the next PR in this series of changes. We split the PRs here because fixing the TODO above ended up requiring some deeper changes that run [through Core](https://github.com/NVIDIA-Merlin/core/pull/280) and several of the other libraries.

merlin/systems/dag/runtimes/nvtabular/executor.py
karlhigley: As much as possible, yes. The ideal end state would be getting rid of this executor entirely 👍🏻

merlin/systems/dag/runtimes/nvtabular/runtime.py
```diff
@@ -0,0 +1,40 @@
...
+NVTABULAR_OP_TABLE = OpTable()
+NVTABULAR_OP_TABLE.register(
+    Categorify, nvtabular_cpp.inference.CategorifyTransform, lambda op: op.encode_type != "combo"
+)
+NVTABULAR_OP_TABLE.register(
+    FillMissing, nvtabular_cpp.inference.FillTransform, lambda op: not op.add_binary_cols
+)
+
+
+class NVTabularServingRuntime(Runtime):
```
oliverholworthy, line 37: From the docstring of the `Runtime` class, this was initially written to run Systems Graphs specifically. I suppose this runtime would work on systems graphs too, but isn't really something you'd pass to the Ensemble. Probably fine to re-use the runtime here for now, but thought it was worth mentioning

Issue #322, opened 2023-04-11
Have a padding operator and add Support for ragged inputs  that we can use consistently with Transformers for rec


viswa-nvidia:

oliverholworthy:
Support for ragged inputs in Transformers4Rec added in https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/666

oliverholworthy:
Tested with NVT workflow with ragged outputs and model with ragged inputs here: https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/677

Using a Padding Operator also works in between NVT Workflow and Transformers4Rec

Issue #135, opened 2022-07-11
[FEA] Add an end-2-end POC example with multi-hot columns


rnyak:
# 🚀 Feature request

We have been asked about an end-to-end example with multi-hot columns. Currently we do not have an end-2-end POC example notebook with multi-hot columns. For such notebook, we can use Movielens 100k dataset, which has user and item features.

karlhigley:
We have support for ragged list columns in Systems now, though there may still be some issues with the ops to work out. The best way to find that out would be to try adding a list column to the multi-stage example notebook and see what issues crop up. I'm sending this back to `examples`. 😄

PR #324, opened 2023-04-12
Support Ragged inputs and outputs in `TransformWorkflow` with Triton

oliverholworthy:
Part of #322 and #135
Replaces #316 

Updates `TransformWorkflow` Triton workflow model to handle NVTabular workflows that have ragged inputs or outputs.

Adds tests for the `TransformWorkflow` usage with ragged inputs and outputs and padded outputs.

### DIFF
--- a/merlin/systems/triton/utils.py
+++ b/merlin/systems/triton/utils.py
@@ -208,9 +208,12 @@ def send_triton_request(
     """
     from merlin.table import TensorTable
 
+    close_client = False
+
     if not client:
         try:
             client = grpcclient.InferenceServerClient(url=endpoint)
+            close_client = True
         except Exception as e:
             raise e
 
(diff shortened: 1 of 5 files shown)


merlin/systems/triton/utils.py
```diff
@@ -223,11 +226,14 @@ def send_triton_request(
         triton_inputs = triton.convert_df_to_triton_input(schema, inputs, grpcclient.InferInput)
 
     outputs = [grpcclient.InferRequestedOutput(col) for col in outputs_list]
-    with client:
-        response = client.infer(triton_model, triton_inputs, request_id=request_id, outputs=outputs)
+
+    response = client.infer(triton_model, triton_inputs, request_id=request_id, outputs=outputs)
```
oliverholworthy, line 230: removing this with is necessary to support calling this function mutliple times with the same client

merlin/systems/workflow/base.py
```diff
@@ -105,7 +109,16 @@ def run_workflow(self, input_tensors):
         if kind != Supports.CPU_DICT_ARRAY:
             transformed, kind = convert_format(transformed, kind, Supports.CPU_DICT_ARRAY)
 
-        return TensorTable(transformed).to_dict()
+        output_table = TensorTable(transformed)
+
+        for col in self.workflow.output_schema:
+            if col.is_ragged and output_table[col.name].offsets is None:
+                values, offsets = _to_ragged(output_table[col.name].values)
+                values = values.astype(self.output_dtypes[col.name + "__values"])
```
oliverholworthy, line 117: adding back the dtype coercion that was removed in #317 to handle the case where some workflows output a dtype that is inconsistent with the `workflow.output_schema`.

karlhigley (approved):

PR #327, opened 2023-04-13
use concurrency setting to kill already-running jobs when new commits come in

nv-alaiacano:
This should help relieve the stress on our CI runners (and also on github's, for the cpu tests).

If this works well, we can put it in all of the other workflows.


nv-alaiacano:
I made a second commit to this PR and confirmed that the first workflow got killed!

karlhigley (approved):

PR #335, opened 2023-04-24
Fix the code that aligns returned data with the output schema

karlhigley:
- `match_representations` was implemented a bit weirdly and seems to have had corner cases we weren't tracking. This updates the implementation to be less clever and more straightforward, which also makes it work better.

### DIFF
--- a/merlin/systems/triton/conversions.py
+++ b/merlin/systems/triton/conversions.py
@@ -83,17 +83,23 @@ def match_representations(schema: Schema, dict_array: Dict[str, Any]) -> Dict[st
     Dict[str, Any]
         A dictionary of NumPy or CuPy ndarrays with representations adjusted
     """
-    schema_names = tensor_names(schema)
-
     aligned = {}
-    for tensor_name in dict_array.keys():
-        if tensor_name in schema_names:
-            aligned[tensor_name] = dict_array[tensor_name]
+    for col_name, col_schema in schema.column_schemas.items():
+        if col_schema.is_ragged:
+            vals_name = f"{col_name}__values"
+            offs_name = f"{col_name}__offsets"
+
+            try:
+                # Look for values and offsets that already exist
+                aligned[vals_name] = dict_array[vals_name]
+                aligned[offs_name] = dict_array[offs_name]
+            except KeyError:
+                # If you don't find them, create the offsets
+                values, offsets = _to_values_offsets(dict_array[col_name])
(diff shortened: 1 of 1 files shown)


jperez999 (approved):

merlin/systems/triton/conversions.py
```diff
@@ -111,12 +111,19 @@ def _to_values_offsets(array):
...
-    num_rows = array.shape[0]
-    row_lengths = [array.shape[1]] * num_rows
-    offsets = [0] + list(itertools.accumulate(row_lengths))
     array_lib = cp if cp and isinstance(array, cp.ndarray) else np
+    shape = array.shape
+
+    values_shape = [-1]
+    if len(shape) > 2:
+        values_shape.extend(shape[2:])
+    values = array.reshape(*values_shape)
```
jperez999, line 120: since we only use values_shape in this call, are we setting it above to make this more readable?

karlhigley: Yeah, it could be an if/else like this:
```python
if len(shape) > 2:
    array.reshape([-1, **shape[2:]])
else:
    array.reshape([-1])
```
That felt repetitive though so I figured I'd just do the reshape once.

merlin/systems/triton/conversions.py
```diff
@@ -111,12 +117,19 @@ def _to_values_offsets(array):
...
+
+    num_rows = shape[0]
+    row_lengths = [shape[1]] * num_rows if len(shape) > 1 else [1]
```
oliverholworthy, line 129: What's the scenario where len(shape) is not going to be greater than 1 when calling this function? Does this [-1] need to be multiplied by num_rows also?

suggests we're converting from 1d array to a ragged list? Another option could be to perform an expand dims in the match representations function if we want to do this, which might be slightly clearer to understand the intent of the function that way?

oliverholworthy:
> match_representations was implemented a bit weirdly and seems to have had corner cases we weren't tracking. This updates the implementation to be less clever and more straightforward, which also makes it work better.


By 'work better', is this about filtering out columns that are not present in the schema from the data passed to match representations?

merlin/systems/triton/conversions.py
karlhigley: On further testing, it looks like these changes aren't actually required to get the example test passing, so I'm removing them from the PR.

PR #344, opened 2023-05-09
update feast op to handle feast 0.31 - latest

jperez999:
This PR updates the feast op to handle the newest version of feast (0.31). Will need to update Merlin repo example in tandem with this PR, to ensure nothing breaks.


oliverholworthy:
Do we also need to update the feast version in `requirements/test.txt`?

nv-alaiacano (approved):

PR #349, opened 2023-05-18
fix readme and requirement files

jperez999:
Fix readme to refer to systems repo when providing feed back and fixed import paths for operators.
Removed feast from gpu requirements as it is not actually required to run systems only necessary if you want to use the feast op
Updated the merlin core and nvtabular dependencies


rnyak:
@jperez999 shall we update this [example nb ](https://github.com/NVIDIA-Merlin/Merlin/blob/main/examples/Building-and-deploying-multi-stage-RecSys/01-Building-Recommender-Systems-with-Merlin.ipynb)as well? It reads as 

we have a cell at the beginning:
```
# for running this example on GPU, install the following libraries
# %pip install "feast<0.20" faiss-gpu
```

shall we remove `"feast<0.20"` ? 

thanks..

PR #357, opened 2023-05-31
[TMP] - Debug Implicit Operator

oliverholworthy:
Placeholder PR for debugging the implicit test that is failing with out-of-memory on GPU


oliverholworthy:
Tried splitting the parametrizeed test out into separate tests for each combination to check if it was something to do with pytest parametrize not cleaning up after each test run. It turns out to still fail like this, so it seems there is some memory being managed in implicit possibly that is not being cleaned up automatically.

PR #363, opened 2023-06-08
Fix Pytorch integration test postmerge test

oliverholworthy:
Fix Pytorch integration test postmerge test

- Extends `LD_LIBRARY_PATH` instead of replacing it completely
  - This was causing an error `CUDA driver version is insufficient for CUDA runtime version`
  - I suspect this might be due to a combination of other CUDA-related enviroment variables that are not passed through to the tox environment.
- Extends `match_representations` type coercion to work with torch tensors

### DIFF
--- a/.github/workflows/postmerge-gpu.yml
+++ b/.github/workflows/postmerge-gpu.yml
@@ -6,7 +6,6 @@ on:
     branches: [main]
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+"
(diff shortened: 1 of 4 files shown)


oliverholworthy:
Postmerge tests succeeded [here](https://github.com/NVIDIA-Merlin/systems/actions/runs/5211279764/jobs/9403392734?pr=363)

jperez999 (approved):

PR #369, opened 2023-06-12
Remove `use_gpu` argument from Implicit model example notebook

oliverholworthy:
Removes the `use_gpu` argument from the Implicit model example notebook (`examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb`).

This is to support running this notebook as an integration test in a CUDA 12 Environment. [Implicit](https://github.com/benfred/implicit) currnently supports CUDA 11.

The model class we're calling should select the GPU version if it can successfully load the GPU accelerated version (currenlty in a CUDA 11 environment)

### DIFF
--- a/examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb
+++ b/examples/Serving-An-Implicit-Model-With-Merlin-Systems.ipynb
@@ -106,8 +106,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ensemble_export_path = os.environ.get(\"OUTPUT_DATA_DIR\", \"ensemble\")\n",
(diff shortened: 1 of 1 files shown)


jperez999 (approved):

jperez999:
This means that in the next release we will have a hard dependency on merlin-models 23.06 right?

oliverholworthy:
> This means that in the next release we will have a hard dependency on merlin-models 23.06 right?

I don't think this PR adds any dependency, [merlin-models is not part of the `install_requires` for Systems](https://github.com/NVIDIA-Merlin/systems/blob/6372e6460ed422c5d6d499943092e014651a14fa/requirements/base.txt). And this PR doesn't change that.

nv-alaiacano (approved):

PR #371, opened 2023-06-21
make the xgboost notebook example runnable

radekosmulski:
This notebook was not running because:

* it was missing `import os` (test was passing because it was import `os` in injected cell)
* for some reason, when run on the movielens synthetic dataset, `convert_df_to_triton_input` was failing where it was not before, I believe

The issue with `convert_df_to_triton_input` was that in the schema of the generated dataset (which aligns with the original) we have the `genres` column marked as rugged, list, whereas the generated column is of type int32 or int64. This mismatch causes `convert_df_to_triton` to fail.

### DIFF
--- a/examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
+++ b/examples/Serving-An-XGboost-Model-With-Merlin-Systems.ipynb
@@ -85,28 +85,35 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "id": "d0385d38",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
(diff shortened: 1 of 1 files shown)


bschifferer (approved):

oliverholworthy:
@radekosmulski please can you provide some context in the description to explain how it makes this notebook runnable. Something wasn't working before?

radekosmulski:
Sorry @oliverholworthy, I discussed this with Benedikt on slack and didn't copy the details here, apologies! Will be more thorough with providing details in my PRs going forward.

radekosmulski:
Sorry @oliverholworthy, I didn't add it to my previous message and you might miss it -- I have now included the details in the description of the PR.

Sorry again for the trouble on this!

PR #372, opened 2023-06-21
Subgraph support changes for multistage example

jperez999:
Added supported format flags for specific operators that may require specific data types internally. Added the capability to load artifacts in a local runtime, was only used previously for triton runtimes after export. Migrated artifact load and save to core https://github.com/NVIDIA-Merlin/core/pull/349. Removed hasattr checks in triton python model class because all operators now have artifact hooks.

### DIFF
--- a/merlin/systems/dag/ops/feast.py
+++ b/merlin/systems/dag/ops/feast.py
@@ -288,7 +288,4 @@ class QueryFeast(InferenceOperator):
 
     @property
     def supported_formats(self) -> DataFormats:
-        return (
-            DataFormats.NUMPY_TENSOR_TABLE
-            | DataFormats.CUPY_TENSOR_TABLE
-        )
+        return DataFormats.NUMPY_TENSOR_TABLE | DataFormats.CUPY_TENSOR_TABLE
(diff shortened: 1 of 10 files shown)


merlin/systems/dag/ops/feast.py
```diff
@@ -285,3 +285,10 @@ def _prefixed_name(cls, output_prefix, col_name):
             return f"{output_prefix}_{col_name}"
         else:
             return col_name
+
+    @property
+    def supported_formats(self) -> DataFormats:
+        return (
+            DataFormats.NUMPY_TENSOR_TABLE
+            | DataFormats.CUPY_TENSOR_TABLE
+        )
```
karlhigley, line 294: Is this not the default for ops in Systems? I see that some ops are flagged as only supporting tensor tables and some as only supporting data frames. Is the default support for all of those?

jperez999: No we never set it as the default in the inference operator https://github.com/NVIDIA-Merlin/systems/blob/main/merlin/systems/dag/ops/operator.py

PR #379, opened 2023-07-06
Remove `InferenceNode` after migrating its functionality into the `Node` class from Merlin Core

jperez999:
This PR removes the inference node and operator and moves the model registry apparatus to core. Now all operators are either base operator or stat operator. This is the first step in making all operators work in workflows and ensembles alike. 

This PR requires the following core PR: https://github.com/NVIDIA-Merlin/core/pull/357


tests/unit/systems/dag/test_model_registry.py
```diff
@@ -36,16 +36,34 @@ def get_artifact_uri(self) -> str:
 
     registry = SimpleModelRegistry()
 
-    # Make a new subclass of InferenceOperator so the mocks don't interfere with other tests.
-    class InferenceOperatorWithModelPath(InferenceOperator):
-        pass
+    # Make a new subclass of BaseOperator so the mocks don't interfere with other tests.
```
karlhigley, line 39: This will make the test pass, but now this functionality only exists in the tests. Unless we can figure out a way to put it the non-test code, we might just want to drop it for now. Writing out the actual code in the method every time you want to do this isn't great, but it's not that much and we can come back to this to make it more convenient once we have a better handle on how model registries should integrate with the rest of Merlin.

karlhigley (approved):

